PR #31, opened 2023-07-27
fix corpus tests

andrewmwells-amazon:
*Issue #, if available:*
https://github.com/cedar-policy/cedar-java/issues/30

*Description of changes:*
Updates to fix the corpus tests. Remove usage of `__expr` escape.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #32, opened 2023-07-28
reenable integration tests

andrewmwells-amazon:
*Issue #, if available:*
https://github.com/cedar-policy/cedar-java/issues/30

*Description of changes:*
In preparation for https://github.com/cedar-policy/cedar-java/issues/30 we updated the integration tests to no longer use the old `__expr` syntax (https://github.com/cedar-policy/cedar/issues/181). This commit fixes CedarJava and reenables integration tests after that change.


CedarJava/src/test/java/com/cedarpolicy/SharedIntegrationTests.java
```diff
@@ -176,25 +176,25 @@ private static class JsonEntity {
...
-//        "tests/ip/1.json",
-//        "tests/ip/2.json",
-//        "tests/ip/3.json",
-//        "tests/multi/1.json",
-//        "tests/multi/2.json",
+        "tests/ip/1.json",
+        "tests/ip/2.json",
+        "tests/ip/3.json",
+        "tests/multi/1.json",
+        "tests/multi/2.json",
```
khieta, line 197: Although these lines are uncommented, the ones below that actually run the tests (L220-242) still appear to be commented out :)

khieta:
Noting for future reference: many of the changes needed to make the corpus tests work were made in #31.

CedarJava/src/test/java/com/cedarpolicy/SharedIntegrationTests.java
andrewmwells-amazon: I think these tests are run. The commented lines are for the corpus tests, which still have some failures.

shaobo-he-aws (approved):

CedarJava/src/test/java/com/cedarpolicy/SharedIntegrationTests.java
khieta: Good point. I shouldn't try to review things on Friday evenings...

CedarJava/src/test/java/com/cedarpolicy/pbt/IntegrationTests.java
```diff
@@ -57,20 +58,20 @@ public void testResourceAttribute() {
         String principal = "User::\"alice\"";
         Map<String, Value> principalAttributes = new HashMap<>();
         Set<String> principalParents = new HashSet<>();
-        Entity e = new Entity(principal, principalAttributes, principalParents);
+        Entity e = new Entity(new JsonEUID("User", "alice"), principalAttributes, principalParents);
```
khieta, line 61: Nit: It feels a little awkward to have "User" and "alice" duplicated in two places (L58 and L61). It might be cleaner if the constructor for `JsonEUID` could take as input a full EUID string and parse it, or if the `principal` string was constructed based on two other variables (one for "User" and one for "alice").

CedarJava/src/test/java/com/cedarpolicy/pbt/IntegrationTests.java
```diff
@@ -197,7 +198,7 @@ public void testRandomResourceAttributeDeny(@ForAll @IntRange(min = 1, max = 50)
         }
         resourceAttributes.put("name", new PrimString("my_photo"));
         Set<String> resourceParents = new HashSet<>();
-        e = new Entity(resource, resourceAttributes, resourceParents);
+        e = new Entity(new JsonEUID("Resource", "my_photo.jpg"), resourceAttributes, resourceParents);
```
khieta, line 201: Per the original code, this should be "photo.jpeg"

khieta (approved):

aaronjeline (approved):

CedarJava/src/test/java/com/cedarpolicy/pbt/IntegrationTests.java
andrewmwells-amazon: I want to encourage use of the two argument version to avoid issues with quotes around `alice`. I don't think we can deprecate the old fn, but I'm trying to avoid using it in tests/examples.

I can save `User` and `alice` into two string and construct `principal` from them

shaobo-he-aws (approved):

PR #43, opened 2023-08-17
Updates to documentation

khieta:
*Issue #, if available:*

*Description of changes:*

* Revised text in the toplevel README to indicate that users should use a `release` branch unless they're building the `cedar-policy` crate locally.
* Updated the `Cargo.toml` in `CedarJavaFFI` to point to a local copy of `cedar` rather than the one available on crates.io.
* Updated links in `bug_report.yml` to point to the correct repo.
* Copied `feature_request.yml` from our [`cedar`](https://github.com/cedar-policy/cedar) repo.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #45, opened 2023-08-18
Don't assume local repo with cedar-policy

andrewmwells-amazon:
Don't assume local repo with cedar-policy

### DIFF
--- a/CedarJavaFFI/Cargo.toml
+++ b/CedarJavaFFI/Cargo.toml
@@ -9,7 +9,7 @@ version = "2.0.0"
 [dependencies]
 serde = { version = "1.0", features = ["derive", "rc"] }
 serde_json = "1.0"
-cedar-policy = { version = "2.3", "../cedar/cedar-policy" } # Need latest version from github
+cedar-policy = { version = "2.3", path = "../cedar/cedar-policy" } # Need latest version from github
 
 # JNI Support
 jni = "0.20.0"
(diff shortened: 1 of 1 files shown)


aaronjeline (approved):

khieta:
@andrewmwells-amazon I think we want to leave this as-is. I updated it in #43 since the `main` branch requires the latest version of `cedar-policy` -- not the version on crates.io. We definitely want the crates.io version in the `release/x.x.x` branches though.

andrewmwells-amazon (commented):
Changed to a comment explaining why we need `path`

andrewmwells-amazon:
> @andrewmwells-amazon I think we want to leave this as-is. I updated it in https://github.com/cedar-policy/cedar-java/pull/43 since the main branch requires the latest version of cedar-policy -- not the version on crates.io. We definitely want the crates.io version in the release/x.x.x branches though.

I see. Not ideal, but probably ok for now. It breaks cedar-java-hello-world, but I think it's reasonable for that to depend on a release instead of `main`. Added a comment and we can fix it after the next `cedar-policy` release.

khieta (approved):

khieta:
> I see. Not ideal, but probably ok for now. It breaks cedar-java-hello-world, but I think it's reasonable for that to depend on a release instead of main. Added a comment and we can fix it after the next cedar-policy release.

We may run into this situation again in the future (un-released breaking changes on the `main` branch in `cedar`) so there may not be a great fix. I think we just want to redirect all users to the release branches (and the released version on Maven, once available).

But this is a bit of a foot gun currently: should we switch to using `release/2.3.x` as the default branch, and rename `main` to `experimental`?

Issue #51, opened 2023-10-04
Support the new EntityUID format for integration tests


aaronjeline:
### Category

Internal refactors/changes

### Describe the feature you'd like to request

`cedar-java` should update to support the new format for integration tests

### Describe alternatives you've considered

N/A

### Is this something that you'd be interested in working on?

- [x] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

khieta:
Closing since this issue is stale. Integration tests are currently working for Java code.

PR #54, opened 2023-10-24
Features/aaronjeline/new euid format

aaronjeline:
*Issue #, if available:*
#51 
*Description of changes:*
Removes all use of `__expr` and reworks the EnittyUID class

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/model/AuthorizationRequest.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/model/AuthorizationRequest.java
@@ -39,7 +39,8 @@ import java.util.Optional;
 public class AuthorizationRequest {
     /** EUID of the principal in the request. */
     @JsonProperty("principal")
-    public final Optional<EntityUID> principalEUID; /** EUID of the action in the request. */
+    public final Optional<EntityUID> principalEUID; 
+    /** EUID of the action in the request. */
     @JsonProperty("action")
     public final EntityUID actionEUID;
     /** EUID of the resource in the request. */
--- a/CedarJava/src/main/java/com/cedarpolicy/model/AuthorizationRequest.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/model/AuthorizationRequest.java
@@ -97,7 +97,7 @@ public class AuthorizationRequest {
     }
 
     /**
-     * Create a request in the empty context.
+     * Create a request without a schema.
      *
      * @param principalEUID Principal's EUID.
      * @param actionEUID Action's EUID.
--- a/CedarJava/src/main/java/com/cedarpolicy/model/AuthorizationRequest.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/model/AuthorizationRequest.java
@@ -17,6 +17,8 @@
 package com.cedarpolicy.model;
 
 import com.cedarpolicy.model.schema.Schema;
+import com.cedarpolicy.model.slice.Entity;
+import com.cedarpolicy.value.EntityUID;
 import com.cedarpolicy.value.Value;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.HashMap;
@@ -37,13 +39,12 @@ import java.util.Optional;
 public class AuthorizationRequest {
     /** EUID of the principal in the request. */
     @JsonProperty("principal")
-    public final Optional<String> principalEUID;
-    /** EUID of the action in the request. */
+    public final Optional<EntityUID> principalEUID; /** EUID of the action in the request. */
     @JsonProperty("action")
-    public final String actionEUID;
+    public final EntityUID actionEUID;
     /** EUID of the resource in the request. */
     @JsonProperty("resource")
-    public final Optional<String> resourceEUID;
+    public final Optional<EntityUID> resourceEUID;
 
     /** Key/Value map representing the context of the request. */
     public final Optional<Map<String, Value>> context;
@@ -61,9 +62,9 @@ public class AuthorizationRequest {
      * @param schema Schema (optional).
      */
     public AuthorizationRequest(
-            Optional<String> principalEUID,
-            String actionEUID,
-            Optional<String> resourceEUID,
+            Optional<EntityUID> principalEUID,
+            EntityUID actionEUID,
+            Optional<EntityUID> resourceEUID,
             Optional<Map<String, Value>> context,
             Optional<Schema> schema) {
         this.principalEUID = principalEUID;
@@ -85,7 +86,7 @@ public class AuthorizationRequest {
      * @param resourceEUID Resource's EUID.
      * @param context Key/Value context.
      */
-    public AuthorizationRequest(String principalEUID, String actionEUID, String resourceEUID, Map<String, Value> context) {
+    public AuthorizationRequest(EntityUID principalEUID, EntityUID actionEUID, EntityUID resourceEUID, Map<String, Value> context) {
         this(
                 Optional.of(principalEUID),
                 actionEUID,
@@ -93,6 +94,34 @@ public class AuthorizationRequest {
                 Optional.of(context),
                 Optional.empty());
     }
+
+    /**
+     * Create a request in the empty context.
+     *
+     * @param principalEUID Principal's EUID.
+     * @param actionEUID Action's EUID.
+     * @param resourceEUID Resource's EUID.
+     * @param context Key/Value context.
+     */
+    public AuthorizationRequest(Entity principalEUID, Entity actionEUID, Entity resourceEUID, Map<String, Value> context) {
+        this(
+                Optional.of(principalEUID.getEUID()),
+                actionEUID.getEUID(),
+                Optional.of(resourceEUID.getEUID()),
+                Optional.of(context),
+                Optional.empty());
+    }
+
+    public AuthorizationRequest(Optional<Entity> principal, Entity action, Optional<Entity> resource, Optional<Map<String, Value>> context, Optional<Schema> schema) {
+        this(
+            principal.map(e -> e.getEUID()),
+            action.getEUID(),
+            resource.map(e -> e.getEUID()),
+            context,
+            schema
+        );
+    }
+
 
     /** Readable string representation. */
     @Override
--- a/CedarJava/src/main/java/com/cedarpolicy/model/slice/Entity.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/model/slice/Entity.java
@@ -16,7 +16,7 @@
 
 package com.cedarpolicy.model.slice;
 
-import com.cedarpolicy.serializer.JsonEUID;
+import com.cedarpolicy.value.EntityUID;
 import com.cedarpolicy.value.Value;
 
 import java.util.*;
@@ -29,13 +29,13 @@ import java.util.stream.Collectors;
  * entities.
  */
 public class Entity {
-    private final JsonEUID euid;
+    private final EntityUID euid;
 
     /** Key/Value attribute map. */
     public final Map<String, Value> attrs;
 
     /** Set of entity EUIDs that are parents to this entity. */
-    public final Set<JsonEUID> parentsEUIDs;
+    public final Set<EntityUID> parentsEUIDs;
 
     /**
      * Create an entity from JsonEUID and unwrapped JSON values.
@@ -44,7 +44,7 @@ public class Entity {
      * @param attributes Key/Value map of attributes.
      * @param parentsEUID Set of parent entities' EUIDs.
      */
-    public Entity(JsonEUID uid, Map<String, Value> attributes, Set<JsonEUID> parentsEUIDs) {
+    public Entity(EntityUID uid, Map<String, Value> attributes, Set<EntityUID> parentsEUIDs) {
         this.attrs = new HashMap<>(attributes);
         this.euid = uid;
         this.parentsEUIDs = parentsEUIDs;
@@ -73,7 +73,7 @@ public class Entity {
      * Get entity uid in JsonEUID format
      * @return Entity UID in JsonEUID format
      */
-    public JsonEUID getEUID() {
+    public EntityUID getEUID() {
         return euid;
     }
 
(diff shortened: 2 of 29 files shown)


CedarJava/src/main/java/com/cedarpolicy/model/AuthorizationRequest.java
```diff
@@ -37,13 +39,12 @@
 public class AuthorizationRequest {
     /** EUID of the principal in the request. */
     @JsonProperty("principal")
-    public final Optional<String> principalEUID;
-    /** EUID of the action in the request. */
+    public final Optional<EntityUID> principalEUID; /** EUID of the action in the request. */
```
cdisselkoen, line 42:
```suggestion
    public final Optional<EntityUID> principalEUID; 
    /** EUID of the action in the request. */
```

CedarJava/src/main/java/com/cedarpolicy/model/AuthorizationRequest.java
```diff
@@ -94,6 +95,34 @@ public AuthorizationRequest(String principalEUID, String actionEUID, String reso
                 Optional.empty());
     }
 
+    /**
+     * Create a request in the empty context.
```
cdisselkoen, line 99: this doesn't seem to be for the empty context.  it's just no schema?

CedarJava/src/main/java/com/cedarpolicy/model/slice/Entity.java
```diff
@@ -73,15 +73,15 @@ public String toString() {
...
+    public EntityUID getEUID() {
         return euid;
     }
 
     /**
      * Get entity uid in JsonEUID format
      * @return Entity UID in JsonEUID format
      */
-    public Set<JsonEUID> getParents() {
+    public Set<EntityUID> getParents() {
```
cdisselkoen, line 84: should the doc comment also no longer refer to JsonEUID?

CedarJava/src/main/java/com/cedarpolicy/model/slice/Entity.java
```diff
@@ -73,15 +73,15 @@ public String toString() {
      * Get entity uid in JsonEUID format
      * @return Entity UID in JsonEUID format
      */
-    public JsonEUID getEUID() {
+    public EntityUID getEUID() {
```
cdisselkoen, line 76: should the doc comment also no longer refer to JsonEUID?

CedarJava/src/main/java/com/cedarpolicy/serializer/JsonEUID.java
```diff
@@ -30,8 +36,14 @@ public class JsonEUID {
     public final String id;
```
cdisselkoen, line 36: should this be `EntityIdentifier`?

CedarJava/src/main/java/com/cedarpolicy/value/EntityTypeName.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    protected EntityTypeName(List<String> namespace, String  basename) {
+        this.namespace = namespace;
+        this.basename= basename;
+    }
+
+    /**
+     * Construct an EntityUID of this type
+     * @param id The EntityIdentifier for the new EntityUID
+     * @return An EntityUID with this time and the supplied EntityIdentifier
```
cdisselkoen, line 37:
```suggestion
     * @return An EntityUID with this type and the supplied EntityIdentifier
```

CedarJava/src/main/java/com/cedarpolicy/value/EntityTypeName.java
```diff
@@ -0,0 +1,116 @@
...
+     */
+    public EntityUID of(String id) {
+        return new EntityUID(this, id);
+    }
+
+    /**
+     * Construct an EntityUID of this type
+     * @param id The EntityIdentifier for the new EntityUID
+     * @return An EntityUID with this time and the supplied EntityIdentifier
```
cdisselkoen, line 46:
```suggestion
     * @return An EntityUID with this type and the supplied EntityIdentifier
```

CedarJava/src/main/java/com/cedarpolicy/value/EntityTypeName.java
```diff
@@ -0,0 +1,116 @@
...
+    public Stream<String> getNamespaceComponents() {
+        return namespace.stream();
+    }
+
+    protected List<String> getNamespace() {
+        return namespace;
+    }
+
+    /** 
+     * Get the namespace component has a string
```
cdisselkoen, line 66:
```suggestion
     * Get the namespace component as a string
```

CedarJava/src/main/java/com/cedarpolicy/value/EntityUID.java
```diff
@@ -17,118 +17,105 @@
...
 
+    /**
+     * Construct an EntityUID from a tyep name and an id
+     * @param type the Entity Type of this EUID
+     * @param id the id portion of the EUID
+     */
+    public EntityUID(EntityTypeName type, EntityIdentifier id) {
+        this.type = type;
+        this.id = id;
+        this.repr = getEUIDRepr(type, id);
```
cdisselkoen, line 44: does this mean we do the native call to get the repr every time we make an EUID? is it needed all the time? or should we be lazy about that?

CedarJava/src/main/java/com/cedarpolicy/value/EntityUID.java
```diff
@@ -17,118 +17,105 @@
...
+    }
 
     /**
-     * Build EntityUID.
-     *
-     * @param euid Entity Unique ID as a string.
-     *     <p>Note, we limit euids to 1024 chars.
+     * Construct an EntityUID from a tyep name and an id
```
cdisselkoen, line 48:
```suggestion
     * Construct an EntityUID from a type name and an id
```

CedarJava/src/main/java/com/cedarpolicy/value/EntityUID.java
```diff
@@ -17,118 +17,105 @@
...
+                return this.type.equals(rhs.type) && this.id.equals(rhs.id);
+            } catch (ClassCastException e) {
+                return false;
+            }
+        }
+    }
+
+    @Override
+    public int hashCode() {
+        return type.hashCode() + id.hashCode();
```
cdisselkoen, line 96: is `+` normal? should we use bitwise-xor instead?

CedarJava/src/test/java/com/cedarpolicy/EntityIdTests.java
```diff
@@ -0,0 +1,8 @@
+import org.junit.jupiter.api.Test;
+
+import com.cedarpolicy.value.EntityIdentifier;
+
+public class EntityIdTests {
+
+    
+}
```
cdisselkoen, line 8: committed by mistake?

CedarJavaFFI/src/interface.rs
```diff
@@ -49,25 +59,25 @@ fn call_cedar_in_thread(call_str: String, input_str: String) -> String {
...
-    let parsing_err = || build_err_obj(env, "parsing");
-    let getting_err = || build_err_obj(env, "getting");
+    let parsing_err = build_err_obj(&env, "parsing");
+    let getting_err = build_err_obj(&env, "getting");
 
-    let j_call_str: String = match env.get_string(j_call) {
+    let j_call_str: String = match env.get_string(&j_call) {
         Ok(call_str) => call_str.into(),
-        _ => return getting_err(),
+        _ => return getting_err,
```
cdisselkoen, line 72: would it be better to just inline the call to `build_err_obj()` here? so we don't construct the error object in Java unless we're actually returning the error?

CedarJavaFFI/src/interface.rs
```diff
@@ -49,25 +59,25 @@ fn call_cedar_in_thread(call_str: String, input_str: String) -> String {
...
     };
-    let j_call_str = j_call_str.clone();
+    let j_call_str = j_call_str;
```
cdisselkoen, line 74: can we remove this?

CedarJavaFFI/src/interface.rs
```diff
@@ -49,25 +59,25 @@ fn call_cedar_in_thread(call_str: String, input_str: String) -> String {
...
 
-    let j_input_str: String = match env.get_string(j_input) {
+    let j_input_str: String = match env.get_string(&j_input) {
         Ok(s) => s.into(),
-        Err(_) => return parsing_err(),
+        Err(_) => return parsing_err,
     };
-    let mut j_input_str = j_input_str.clone();
+    let mut j_input_str = j_input_str;
```
cdisselkoen, line 80: can we remove this?

CedarJavaFFI/src/interface.rs
```diff
@@ -151,6 +176,123 @@ pub fn json_parse_entity_uid(input: &str) -> InterfaceResult {
...
+    match get_entity_identifier_repr_internal(&mut env, obj) {
+        Ok(v) => v.as_jni(),
+        Err(e) => jni_failed(&mut env, e.as_ref()),
+    }
+}
+
+fn get_entity_identifier_repr_internal<'a>(
+    env: &mut JNIEnv<'a>,
+    obj: JObject<'a>,
+) -> Result<JValueOwned<'a>> {
```
cdisselkoen, line 190: for my own edification, what is `JValueOwned` vs `JValueGen` and why do some of these helpers return one vs the other?

CedarJavaFFI/src/objects.rs
```diff
@@ -0,0 +1,311 @@
...
+
+use cedar_policy::{EntityId, EntityTypeName, EntityUid};
+use jni::{
+    objects::{JObject, JString, JValueGen, JValueOwned},
+    sys::jvalue,
+    JNIEnv,
+};
+
+/// General trait for anything that's a wrapper around a  java object
+/// This let's us dynamically cast from a Java Object to our typed wrapper
```
cdisselkoen, line 15:
```suggestion
/// General trait for anything that's a wrapper around a java object.
/// This lets us dynamically cast from a Java Object to our typed wrapper
```

CedarJavaFFI/src/utils.rs
```diff
@@ -0,0 +1,67 @@
...
+    if env.is_same_object(&expected_class, &class)? {
+        Ok(())
+    } else {
+        let class_name = get_class_name(env, class)?;
+        let expected_class_name = get_class_name(env, expected_class)?;
+        env.throw_new(
+            "java/lang/ClassCastException",
+            format!("{class_name} cannot be cast to {expected_class_name}"),
+        )?;
+        Err(Box::new(InternalJNIError::TypeErorr {
```
cdisselkoen, line 27:
```suggestion
        Err(Box::new(InternalJNIError::TypeError {
```

CedarJavaFFI/src/utils.rs
```diff
@@ -0,0 +1,67 @@
...
+#[derive(Debug, Error)]
+pub enum InternalJNIError {
+    #[error("Internal invariant violated, expected member of type `{0}`")]
+    BadMemberType(&'static str),
+    #[error("Internal invariant violated. Object passed to jni function was of the wrong class. Expected: `{expected}`, got: `{got}`")]
+    TypeErorr { expected: String, got: String },
+    #[error("Null pointer")]
+    NullPointer,
+    #[error("Index out of bounds")]
+    IndexOutOfBounds,
```
cdisselkoen, line 52: what the index was and what the bound was?

CedarJavaFFI/src/utils.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/// Given a Java value, extracts the object reference if it exists, otherwise errors
+pub fn get_object_ref(v: JValueGen<JObject<'_>>) -> Result<JObject<'_>> {
+    match v {
+        JValueGen::Object(o) => Ok(o),
+        _ => Err(Box::new(InternalJNIError::BadMemberType("String"))),
```
cdisselkoen, line 59: why "String"?

CedarJava/src/main/java/com/cedarpolicy/value/EntityUID.java
aaronjeline: I was trying avoid needing to redo the native call for every call to `toString()` without adding state to the object, but I think I've gone down the rate of pre-mature op, and should just do the native call on every `toString()`

CedarJavaFFI/src/interface.rs
aaronjeline: `JValueOwned` is a typedef for `JValueGen`, which can represent objects both owned and borrowed

andrewmwells-amazon (approved):

CedarJava/src/main/java/com/cedarpolicy/serializer/JsonEUID.java
andrewmwells-amazon: Don't need to change in this PR, but I think it might now be possible to remove this class? If not, I think it would make sense to store the EntityUID instead of the `type` and `id` separately.

CedarJava/src/main/java/com/cedarpolicy/value/EntityUID.java
```diff
@@ -17,118 +17,104 @@
...
+
+import com.cedarpolicy.serializer.JsonEUID;
 
 /**
  * Represents a Cedar Entity UID. An entity UID contains both the entity type and a unique
  * identifier for the entity formatted as <code>TYPE::"ID"</code>.
  */
-public class EntityUID extends Value {
-
-    private static class EUIDValidator {
```
andrewmwells-amazon, line 30: I think your code is obviously better, but should probably get Barbara to sign off on removing this.

andrewmwells-amazon (approved):

CedarJava/src/main/java/com/cedarpolicy/serializer/JsonEUID.java
aaronjeline: I think so as well but wanted to leave that for a larger refactoring

cdisselkoen (approved):

PR #72, opened 2023-12-15
feat: policy, template, and schema parsing

Swolebrain:
*Issue #, if available:*

n/a

*Description of changes:*

* Align behavior of JEntityTypeName with JEntityId. Previously, the FFI code had a difference in behavior between `JEntityTypeName` and `JEntityId`. The old code did not allow construction of an invalid id, but it did allow construction of an invalid `JEntityTypeName`. I changed it so it fails at construction time if it's invalid.
* Also implemented the same pattern for JEntityUID, and gave it a method `get_rust_repr`.
* New schema parse functionality
* New policy parse functionality
* New functionality to validate a template link - takes in a template and two EntityUIDs.


CedarJava/src/main/java/com/cedarpolicy/model/schema/Schema.java
```diff
@@ -27,6 +28,10 @@
 public final class Schema {
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
+    static {
```
lucajone, line 31: I notice that we now load the library in a few places. I don't know if this ends up slow at all or if it notices it's already been loaded, but if nothing else it might be nice to keep the logic in one place. For example, if we end up doing something more sophisticated like packaging built Cedar libraries in the JAR, it would be nice to only have to replace it once.

Perhaps we could move the JNI definitions to a single package-private class? (Or have a flag somewhere...)

CedarJava/src/main/java/com/cedarpolicy/model/slice/Policy.java
```diff
@@ -41,7 +54,8 @@ public Policy(
             throw new NullPointerException("Failed to construct policy from null string");
         }
         if (policyID == null) {
-            throw new NullPointerException("Failed to construct policy with null ID");
+            policyID = "policy" + idCounter;
+            idCounter += 1;
```
lucajone, line 58: I think this could cause duplicate IDs if we construct `Policy` objects from multiple threads. I guess we could:
  * document this and leave it as-is
  * use an `AtomicInteger` for the counter (presumably the overhead here isn't massive)
  * allow `null` policy IDs in this class – I think the Rust library's constructors now take `Option`s for policy IDs

CedarJava/src/main/java/com/cedarpolicy/model/slice/Policy.java
```diff
@@ -51,4 +65,38 @@ public Policy(
...
+     * Hence we return false, since there's only one failure reason that can't be determined from other means.
+     * @param p Policy object constructed from a valid template. Best if built from Policy.parsePolicyTemplate
+     * @param principal EntityUid to put into the principal slot. Leave null if there's no principal slot
+     * @param resource EntityUid to put into the resource slot. Leave null if there's no resource slot
+     * @return
+     */
+    public static boolean validateTemplateLinkedPolicy(Policy p, EntityUID principal, EntityUID resource) {
+        try {
+            return validateTemplateLinkedPolicyJni(p.policySrc, principal, resource);
+        } catch (Exception e) {
```
lucajone, line 93: Do we want to swallow every exception here? Internal exceptions might not mean an invalid policy.

CedarJavaFFI/src/interface.rs
```diff
@@ -171,6 +171,145 @@ pub fn json_parse_entity_uid(input: &str) -> InterfaceResult {
...
+    template_jstr: JString<'a>,
+    jprincipal_euid: JObject<'a>,
+    jresource_euid: JObject<'a>,
+) -> Result<JValueOwned<'a>> {
+    if template_jstr.is_null() {
+        raise_npe(env)
+    } else {
+        let template_jstring = env.get_string(&template_jstr)?;
+        let template_string = String::from(template_jstring);
+        let template = Template::from_str(&template_string)?;
```
lucajone, line 287: In particular (re. the exception comment above), should we handle these `Err`s as user errors and return false?

CedarJava/src/main/java/com/cedarpolicy/model/schema/Schema.java
Swolebrain: Wouldn't such a class get really big? And wouldn't this be a one-time cost only paid at initialization time?

CedarJava/src/main/java/com/cedarpolicy/model/slice/Policy.java
Swolebrain: What else could they mean? Bear in mind that the principal, resource, and policy have already been parsed by the library so they're valid

CedarJavaFFI/src/interface.rs
Swolebrain: This is an artifact of the cross-language boundary. In order to avoid this i'd have to write a whole other series of JPolicy objects representing the rust-side representations of java heap objects.

CedarJava/src/main/java/com/cedarpolicy/model/schema/Schema.java
lucajone: It is a one-time cost per class. The only thing I could imagine being slightly annoying is if you say pre-construct an `AuthorizationEngine` when your application starts, but then touch the `Schema` or `Policy` classes for the first time in a request context. If calling `loadLibrary` does end up being expensive, then it wouldn't be ideal for this to happen there.

I think something like this would work actually, rather than having to centralise the `native` declarations:

```java
public final class Schema {
  static { CedarLibrary.init(); }
}

static final class CedarLibrary {
  static { System.loadLibrary(...); }

  private CedarLibrary() {}
  public static void init() {}
}
```

...but you're right, this is probably premature optimisation.

CedarJava/src/main/java/com/cedarpolicy/model/slice/Policy.java
lucajone: I'm more thinking of exceptions that don't arise from the policy parsing calls specifically. For example, if we have a bug in our JNI code, perhaps that would surface as an exception from the JVM that we would want to bubble up. Possibly worrying too much here, but if we have a base class for our exceptions sitting around (`InternalException`?) then it might make sense to use that.

CedarJava/src/main/java/com/cedarpolicy/model/schema/Schema.java
Swolebrain: Good point that the Schema class won't actually get the static initialization until you call `new Schema()` somewhere..

CedarJava/src/main/java/com/cedarpolicy/model/slice/Policy.java
Swolebrain: I think im convinced to make it throw

CedarJavaFFI/src/objects.rs
```diff
@@ -38,6 +39,12 @@ impl<'a> JEntityTypeName<'a> {
         basename: JString<'a>,
         namespace: List<'a, JString<'a>>,
     ) -> Result<Self> {
+        let jstr_basename = env.get_string(&basename)?;
+        let basename_str = String::from(jstr_basename);
+        let mut full_type_name: Vec<String> = jstr_list_to_rust_vec(env, &namespace)?;
+        full_type_name.push(basename_str);
+        let fulll_ns_str: String = full_type_name.join("::");
```
andrewmwells-amazon, line 46:
```suggestion
        let full_ns_str: String = full_type_name.join("::");
```

andrewmwells-amazon: We may also want to validate that none of the strings in `namespace` contain `::`.

CedarJavaFFI/src/objects.rs
```diff
@@ -122,8 +107,17 @@ impl<'a> JEntityTypeName<'a> {
...
-        Ok(Self { obj })
+        let namespace = env.call_method(&obj, "getNamespace", "()Ljava/util/List;", &[])?;
+        let namespace = List::cast_unchecked(get_object_ref(namespace)?, env)?;
+        let basename = env.call_method(&obj, "getBaseName", "()Ljava/lang/String;", &[])?;
+        let basename = JString::cast(env, get_object_ref(basename)?)?;
+        let jstr_basename = env.get_string(&basename)?;
+        let basename_str = String::from(jstr_basename);
+        let mut full_type_name: Vec<String> = jstr_list_to_rust_vec(env, &namespace)?;
+        full_type_name.push(basename_str);
+        let fulll_ns_str: String = full_type_name.join("::");
```
andrewmwells-amazon, line 118:
```suggestion
        let full_ns_str: String = full_type_name.join("::");
```

andrewmwells-amazon: Same comment about checking for `::`'s

CedarJava/src/main/java/com/cedarpolicy/model/slice/Policy.java
```diff
@@ -51,4 +65,33 @@ public Policy(
...
+    }
+
+    /**
+     * This method takes in a Policy and a list of Instantiations and calls Cedar JNI to ensure those slots
+     * can be used to instantiate the template. If the Template is validated ahead of time by using Policy.parsePolicyTemplate
+     * and the Instantiations are also ensured to be valid (for example, by validating their parts using EntityTypeName.parse
+     * and EntityIdentifier.parse), then this should only fail because the slots in the template don't match the instantiations
+     * (barring JNI failures).
+     * @param p Policy object constructed from a valid template. Best if built from Policy.parsePolicyTemplate
+     * @param principal EntityUid to put into the principal slot. Leave null if there's no principal slot
```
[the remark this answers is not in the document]
Swolebrain: Optional is not really a fit though, we need to surface why it failed if it fails. We have two choices that offer library ergonomics that are reasonable to java programmers: 
* Just throw exceptions
* Use a library that implements an either monad, or implement it ourselves

CedarJava/src/test/java/com/cedarpolicy/EntityTypeNameTests.java
```diff
@@ -36,7 +36,8 @@ public class EntityTypeNameTests {
         "else",
         "in",
         "like",
-        "has"
+        "has",
+        "is"
```
Swolebrain, line 40: luckily enough i ran into a test failure from this lol

CedarJava/src/main/java/com/cedarpolicy/model/slice/Policy.java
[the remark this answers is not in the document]
andrewmwells-amazon: Sorry, I should have been more clear. I'd prefer that the `principal` and `resource` be `Optional<EntityUID>` instead of using `null` when there's no slot. I agree we should throw an exception on failure, not return an `Optional<boolean>`.

CedarJava/src/test/java/com/cedarpolicy/EntityTypeNameTests.java
andrewmwells-amazon: While it is nice that it will "eventually" catch the issue, but I really don't like non-determinism in CI. Created issue here: https://github.com/cedar-policy/cedar-java/issues/73

andrewmwells-amazon (approved):

CedarJavaFFI/src/objects.rs
```diff
@@ -122,8 +111,21 @@ impl<'a> JEntityTypeName<'a> {
...
+        let jstr_basename = env.get_string(&basename)?;
+        let basename_str = String::from(jstr_basename);
+        let mut full_type_name: Vec<String> = jstr_list_to_rust_vec(env, &namespace_components)?;
+        full_type_name.push(basename_str);
+        let has_namespace_component_with_colon = full_type_name.iter().any(|s| s.contains("::"));
+        if has_namespace_component_with_colon {
+            return Err("components of the type name cannot contain colons".into());
+        }
+        let full_ns_str: String = full_type_name.join("::");
+        let type_name: EntityTypeName = full_ns_str.parse()?;
```
khieta, line 127: This looks like the same code as L42-51. Might be nice to centralize this logic in case we end up changing handling in the future.

khieta (approved):
Not a Java expert, but the new functionality looks good to me.

Issue #79, opened 2024-01-26
Java integration test fails when executed against a release build of the FFI library


ygrignon-sfdc:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar-java/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar-java/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Other

### Describe the bug

If I build CedarJavaFFI using `cargo build` then running `CedarJava/gradlew test` completes successfully.
If I build CedarJavaFFI using `cargo build -r` then running `CedarJava/gradlew test` errors out with a single test failure:

### Expected behavior

Running tests should succeed with both debug and release builds of CedarJavaFFI

### Reproduction steps

1. cd CedarJavaFFI
2. cargo build
3. cd ../CedarJava
4. make sure your Java version is openjdk 17
5. make sure build.gradle's environment 'CEDAR_JAVA_FFI_LIB' entry points to your debug build created by step 2
6. ./gradlew clean test
7. observe that the tests completed successfully
8. cd ../CedarJavaFFI
9. cargo build -r
10. cd ../CedarJava
11. update build.gradle's environment 'CEDAR_JAVA_FFI_LIB' entry to point to the release build created by step 9
12. ./gradlew clean test
13. observe that the tests failed

### Code Snippet

N/A

### Log output

```
$ ./gradlew clean test

> Task :compileTestJava
Note: /Users/ygrignon/projects/cedar-policy/cedar-java/CedarJava/src/test/java/com/cedarpolicy/JSONTests.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: /Users/ygrignon/projects/cedar-policy/cedar-java/CedarJava/src/test/java/com/cedarpolicy/ValidationTests.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :test

IntegrationTests > testLongExprStackOverflowDeny() FAILED
    org.opentest4j.AssertionFailedError: expected: <false> but was: <true>
        at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
[...]

PR #80, opened 2024-01-29
Ensures our test for stack overflows correctly triggers a stack overflow on release build

aaronjeline:
*Issue #, if available:*
#79 
*Description of changes:*
We have an integration tests that ensures the correct behavior for stack overflows in the rust interpreter. 
Because rust stack frames are much smaller in release mode, the test failed to trigger a stack overflow when run under release mode.
This just makes the test case larger so that we trigger a stack overflow on both build types.


ygrignon-sfdc (approved):

khieta (approved):
This change looks fine to me. I see that the related issue (#79) is closed. Is this PR still supposed to be open?

andrewmwells-amazon (approved):

Issue #86, opened 2024-02-15
Consider removing Guava dependecy 


john-h-kastner-aws:
### Category

Internal refactors/changes

### Describe the feature you'd like to request

We use Guava for a few utility functions, but this pulls in a relativity larger dependency for some functions we could reasonably implement our selves.

When making this change, consider the differences between the builtin unmodifiable sets/maps and Guava's immutable data structures. In particular, the builtin types can be mutated by mutating a reference to the underlying set.

Previously removed in #81, but restored in #83.

### Describe alternatives you've considered

.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #88, opened 2024-02-19
Re-remove dependency on Guava

john-h-kastner-aws:
Requires the following changes:

* Use `copyOf` and `unmodifiableSet` functions to replace `ImmutableSet`. Prefer using `copyOf` because it creates an unmodifiable copy of the set or list, so mutation to the input does not reflect in the new object. `unmodifiableSet` can still be used to return an immutable view of a set without causing any copying.

* Roll our own memoizing supplier. This change should be reviewed by someone with more Java knowledge, but, AFAIK, the behavior should be the same.

*Issue #, if available:*

#86 

*Description of changes:*

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/value/EntityTypeName.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/value/EntityTypeName.java
@@ -1,8 +1,6 @@
 
 package com.cedarpolicy.value;
 
-import com.google.common.base.Suppliers;
-
 import java.util.List;
 import java.util.Optional;
 import java.util.Objects;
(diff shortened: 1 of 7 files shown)


andrewmwells-amazon (approved):

CedarJava/src/main/java/com/cedarpolicy/value/EntityTypeName.java
```diff
@@ -34,7 +32,15 @@ public final class EntityTypeName {
     protected EntityTypeName(List<String> namespace, String  basename) {
         this.namespace = namespace;
         this.basename= basename;
-        this.entityTypeNameRepr = Suppliers.memoize(() -> getEntityTypeNameRepr(this));
+        this.entityTypeNameRepr = new Supplier<String>() {
```
john-h-kastner-aws, line 35: If anyone is wondering why this is stalled: The Guava memoizing supplier is thread safe, but this implementation would not be. We don't make an explicit guarantees about thread safety of the Java bindings, but out Rust implementation _is_ thread safe, so I don't want to break that here, and I haven't had the time to do this properly.

exceptionfactory:
@john-h-kastner-aws Raising the question, is this change planned for reconsideration at some point? It would be helpful to remove the 3 MB dependency on Google Guava for those integrating with the Cedar Java library.

PR #92, opened 2024-02-20
Update BasicAuthorizationEngine.java to fix logger error

ChadAtAlbanero:
No matching issue

Removed superfluous log statement template placeholder that causes an IllegalArgumentException from BasicAuthorizationEngine::call. It appears that the full request log was moved to the subsequent LOG.trace statement, but the corresponding placeholder in this line was not removed. This can cause authorization requests to fail if the log level is debug or lower. This fix is the same as [PR #90 ](https://github.com/cedar-policy/cedar-java/pull/90), but targets the main branch instead of the current MavenCentral release branch.


khieta (approved):

john-h-kastner-aws (approved):

aaronjeline (approved):
Seems good, although any issues with just dropping the logs here?

andrewmwells-amazon (approved):

Issue #28, opened 2023-07-14
Package native library files in Maven release


andrewmwells-amazon:
### Category

Other

### Describe the feature you'd like to request

It would be nice if we packaged native library files in the Maven release so people don't have to download them from github.

### Describe the solution you'd like

It looks like we could follow https://stackoverflow.com/a/10071519 to include the native libraries in the Maven release.

### Describe alternatives you've considered

Current system.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

exceptionfactory:
@andrewmwells-amazon I am interested in adding native library bundling to `cedar-java`, and I would be glad to put together a pull request if that would be helpful.

In terms of supported operating systems and architectures, what should be prioritized for initial implementation? Starting with Linux and x86_64/amd64 seems like a minimum requirement. The popularity of macOS and aarch64/arm64 is also worth considering. Should supporting Windows be included in the initial scope?

As far as runtime implementation, the Stack Overflow discussion has a helpful example of direct code that reads JVM System properties and loads the library from a preset path in the JAR. The [Fizzed Java Native Environment](https://github.com/fizzed/jne) library also seems worth considering, as it is licensed under Apache-2.0, only depends on SLF4J API, and provides a simple API with convention-based storage of libraries to be loaded. Is there a preferred approach in terms of direct implementation versus such a library?

The last major question relates to the build approach itself. Regardless of the build strategy, the end result requires bundling the platform-specific binaries from a common location. GitHub Actions provides runners for Linux, macOS, and Windows, and just recently introduced macOS 14 with M1 processors, which would allow building for aarch64. It seems like it would be helpful to have a build process that also works a local machine, provided certain required components are installed. This could be implemented through a shell script, or integrated more directly as a task in the Gradle build. Part of the challenge includes how to orchestrate the multi-platform builds for the native Rust library.

To help move this forward, I created a branch with that uses the Fizzed JNE library and builds a JAR that contains both Linux x86_64 and macOS aarch64 binaries using Maven to execute Cargo build commands, running on a GitHub workflow with macOS 14. The following commit contains the initial set of changes:

https://github.com/exceptionfactory/cedar-java/commit/be785ca7e672463f9b511f571035cfb26bde826a

[...]

PR #104, opened 2024-03-13
Add Gradle tasks for bundling native libraries

exceptionfactory:
*Issue #, if available:*

Issue #28

*Description of changes:*

# Summary

This pull request incorporates both code and build changes to provide an uber JAR that bundles the native Cedar Foreign Function Interface libraries required for JNI operations.

# Code Changes

The code changes consists of a new `LibraryLoader` class that introduces a shared method named `loadLibrary()` to handle native library loading. The `LibraryLoader.loadLibrary()` call replaces all instances of `System.load()`, maintaining the existing option to load the native library from a path specified in the `CEDAR_JAVA_FFI_LIB` environment variable.

When the `CEDAR_JAVA_FFI_LIB` environment variable in not specified, `LibraryLoader.loadLibrary()` delegates to the [Java Native Environment](https://github.com/fizzed/jne) `JNE.loadLibrary()` method, which detects the runtime architecture, copies the applicable library to a temporary directory, and calls the appropriate System method. JNE is licensed under Apache Software License 2.0 and provides a conventional approach to native library loading. The standard convention requires that native libraries be placed in a JAR under a `jne` directory with operating system and architecture sub-directories.

# Build Changes

The build changes consist of Gradle tasks that run Cargo build commands and copy native libraries to standard directories. The new Gradle `uberJar` task produces a JAR containing both the standard Java classes and native libraries for the following targets:

- `aarch64-apple-darwin`
- `aarch64-unknown-linux-gnu`
- `x86_64-apple-darwin`
- `x86_64-pc-windows-gnu`
- `x86_64-unknown-linux-gnu`

These targets cover most common uses on Linux, macOS, and Windows for both [AArch64](https://en.wikipedia.org/wiki/AArch64) and [x86_64](https://en.wikipedia.org/wiki/X86-64).

The uber JAR is about 22 MB, which is significant, but provides broad compatibility and includes notable compression for the native libraries binaries.

The Gradle build continues to produce a standard JAR without any native libraries.

The Gradle build tasks use [cargo-zigbuild](https://github.com/rust-cross/cargo-zigbuild) which depends on [Zig](https://ziglang.org/) for cross-compilation. This requires Zig to be installed in addition to Rust, but it enables compilation on either Linux or macOS regardless of version or architecture. This approach allows both local development and GitHub Runners to use the same Gradle tasks without conditional handling in the build configuration. 

The Gradle build updates add the [Maven Publish Plugin](https://docs.gradle.org/current/userguide/publishing_maven.html) to enable local builds using the `./gradlew publishToMavenLocal` command. The Maven configuration properties reflect the values published to Maven Central for [cedar-java](https://central.sonatype.com/artifact/com.cedarpolicy/cedar-java).

Additional changes include upgrading the SpotBugs build plugin from 5.0.14 to 6.0.8 and adding several `SuppressFBWarnings` annotations required as the result of new SpotBugs rules.

# Additional Steps

The goal of this pull request is to introduce support for building a JAR that includes native libraries. Subsequent changes may be required for publishing the new uber JAR to Maven Central with the `uber` classifier.

Next steps could also include building JARs scoped a specific architecture or operating system. However, for projects planning to build and deploy across different platforms, the uber JAR is likely to be the most useful.

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/serializer/JsonEUID.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/serializer/JsonEUID.java
@@ -57,7 +57,7 @@ public class JsonEUID {
         this.type = type; this.id = id;
     }
 
-    @SuppressFBWarnings
+    @SuppressFBWarnings("CT_CONSTRUCTOR_THROW")
     public JsonEUID(String src) throws InvalidEUIDException {
         var o = EntityUID.parse(src);
         if (o.isPresent()) {
--- a/CedarJava/src/main/java/com/cedarpolicy/serializer/JsonEUID.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/serializer/JsonEUID.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.cedarpolicy.value.EntityUID;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /** Represent JSON format of Entity Unique Identifier. */
 @JsonDeserialize
@@ -56,6 +57,7 @@ public class JsonEUID {
         this.type = type; this.id = id;
     }
 
+    @SuppressFBWarnings
     public JsonEUID(String src) throws InvalidEUIDException {
         var o = EntityUID.parse(src);
         if (o.isPresent()) {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,47 +2,30 @@ name: Build & Test (CedarJava & CedarJavaFFI)
 
 on:
   pull_request:
+  workflow_dispatch:
 
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  build_and_test_cedar_java_ffi:
-    name: Rust project - latest
+  build:
+    name: Build FFI and Java
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        toolchain:
-          - stable
     steps:
-      - name: Checkout CedarJava
-        uses: actions/checkout@v3
-      - name: Checkout cedar
-        uses: actions/checkout@v3
-        with:
-          repository: cedar-policy/cedar
-          ref: main
-          path: ./cedar
-      - name: rustup
-        run: rustup update ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
-      - name: cargo fmt
+      - name: Checkout cedar-java
+        uses: actions/checkout@v4
+      - name: Prepare Rust Build
+        run: rustup update stable && rustup default stable
+      - name: Check FFI Formatting
         working-directory: ./CedarJavaFFI
         run: cargo fmt --all --check
-      - name: configure
+      - name: Install Zig
+        run: sudo snap install zig --beta --classic
+      - name: Build FFI and Java Libraries
         working-directory: ./CedarJava
-        shell: bash
-        run: bash config.sh run_int_tests
-      - name: cargo build
-        working-directory: ./CedarJavaFFI
-        run: RUSTFLAGS="-D warnings -F unsafe-code" cargo build --verbose
-      - name: cargo test
-        working-directory: ./CedarJavaFFI
-        run: RUSTFLAGS="-D warnings -F unsafe-code" cargo test --verbose
-      - name: Build and Test CedarJava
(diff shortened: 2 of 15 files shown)


andrewmwells-amazon (approved):
Looks good, thanks for the PR! 

We can merge now, but before we release the shared libraries on Maven we'll have to add a `THIRD_PARTY_LICENSES.txt` file and have AWS's legal department verify it looks ok. So there will be a delay before we push this to Maven Central.

exceptionfactory:
> Looks good, thanks for the PR!
> 
> We can merge now, but before we release the shared libraries on Maven we'll have to add a `THIRD_PARTY_LICENSES.txt` file and have AWS's legal department verify it looks ok. So there will be a delay before we push this to Maven Central.

Thanks for the review @andrewmwells-amazon!

JNE is licensed under Apache Software License 2.0, so hopefully that helps simplify the review process. However, if it is very involved, I believe it would be possible to put together a custom solution that would perform platform detection and load the appropriate library. It isn't trivial, but the it shouldn't be too complicated either. If that approach is preferable, just let me know, otherwise, I look forward to seeing this move forward.

Thanks again!

CedarJava/src/main/java/com/cedarpolicy/serializer/JsonEUID.java
```diff
@@ -56,6 +57,7 @@ public JsonEUID(String type, String id) {
         this.type = type; this.id = id;
     }
 
+    @SuppressFBWarnings
```
lucajone, line 60: Nit: if there are specific new FindBugs warnings that we don't agree with and want to suppress, can we include the particular ones in the annotation? Just so we don't end up masking useful ones which might be introduced later.

lucajone:
Looks good to me, with one minor comment about FindBugs which could be fixed in a followup I think. Thanks for putting this together!

A reminder as well though that ideally we should run the integration tests on each platform we package for before publishing to Maven – especially for macOS I don't know if there are potential signing issues. I don't necessarily mean in CI, we could do it manually (happy to help where I can), although that would be great to have too at some point!

CedarJava/src/main/java/com/cedarpolicy/serializer/JsonEUID.java
john-h-kastner-aws: For reference, the warning is due to throwing exceptions in constructors:

```
M B CT: Exception thrown in class com.cedarpolicy.model.slice.Policy at new com.cedarpolicy.model.slice.Policy(String, String) will leave the constructor. The object under construction remains partially initialized and may be vulnerable to Finalizer attacks.  At Policy.java:[line 55]
```

There's a separate PR to update the findbugs dependency we've been holding off on merging until we've decided what we want to do.

exceptionfactory: Thanks for feedback, I will push an update to list the specific rule for now.

exceptionfactory: The other pull request is from dependabot and does not include the required constructor changes. If you prefer that I back out the SpotBugs upgrade and annotations, that is certainly an option.

exceptionfactory: @lucajone and @john-h-kastner-aws I pushed a commit adding the specific SpotBugs rules to the suppress annotations.

If you would prefer that I back out the SpotBugs plugin upgrade from this PR, I can make adjustments.

exceptionfactory:
> Looks good to me, with one minor comment about FindBugs which could be fixed in a followup I think. Thanks for putting this together!
> 
> A reminder as well though that ideally we should run the integration tests on each platform we package for before publishing to Maven – especially for macOS I don't know if there are potential signing issues. I don't necessarily mean in CI, we could do it manually (happy to help where I can), although that would be great to have too at some point!

Thanks for the feedback @lucajone!

It is worth noting that I did not include any changes to the Publish Release GitHub workflows, as it was not quite clear from the listed steps how the Maven publishing works. With the changes in this pull request, additional Gradle changes could be added to support artifact signing and automated releases. I kept the changes scoped to the local build, which should continue support other release processes, while providing the option for additional automation if desired.

lucajone:
I agree it makes sense to keep the changes scoped as they are. My comments are mostly a note for when we do release those artifacts next that we'll have extra testing steps to do.

shaobo-he-aws (approved):

CedarJava/src/main/java/com/cedarpolicy/serializer/JsonEUID.java
john-h-kastner-aws: We're all happy with it.

PR #107, opened 2024-03-15
Update build process of CedarJavaFFI

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/CedarJavaFFI/Cargo.toml
+++ b/CedarJavaFFI/Cargo.toml
@@ -7,11 +7,10 @@ edition = "2021"
 version = "4.0.0"
 
 [dependencies]
-cedar-policy = { version = "4.0.0", path = "../../cedar/cedar-policy" }
 serde = { version = "1.0", features = ["derive", "rc"] }
 serde_json = "1.0"
-thiserror = "1.0.49"
-itertools = "0.11.0"
+thiserror = "1.0"
+itertools = "0.12"
 
 # JNI Support
 jni = "0.21.0"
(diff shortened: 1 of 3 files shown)


CedarJavaFFI/Cargo.toml
```diff
@@ -23,3 +22,8 @@ partial-eval = ["cedar-policy/partial-eval"]
 # Dynamic linked library, used for linked to java process at runtime.
 [lib]
 crate_type = ["cdylib"]
+
+[dependencies.cedar-policy]
+version = "4.0.0"
+git = "https://github.com/cedar-policy/cedar"
+branch = "main"
```
cdisselkoen, line 29: Instead of this block, why not the equivalent

```
cedar-policy = { version = "4.0.0", git = "https://github.com/cedar-policy/cedar", branch = "main" }
```
in `[dependencies]`?

I think we have another repo that uses this block specifically because there's some GitHub Actions hackery that involves fiddling with Cargo.toml, but are we doing that here?

cdisselkoen (approved):

CedarJavaFFI/Cargo.toml
shaobo-he-aws: I don't think so. These hackeries were introduced because we didn't have proper version handling.

khieta: @cdisselkoen I think the equivalent would be fine here. We can adjust in a future rev.

PR #112, opened 2024-03-22
actually run junit tests

andrewmwells-amazon:
*Issue #, if available:*

*Description of changes:*
The junit tests weren't being run because we were missing `testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.10.2'`

Also enable `partial-eval` in CedarJavaFFI so those tests pass.


khieta (approved):

CedarJava/build.gradle
```diff
@@ -187,10 +178,20 @@ tasks.register('uberJar', Jar) {
 }
 
 tasks.named('test') {
+    useJUnitPlatform()
     dependsOn('compileFFI')
     classpath += files(layout.buildDirectory.dir(compiledLibDir))
 }
 
+test {
+    //environment "CEDAR_INTEGRATION_TESTS_ROOT", ''set to absolute path of `cedar-integration-tests`'
```
khieta, line 187: Ok to delete this line? We already mention that the env variable needs to be set in the readme

khieta (approved):
🎉

.github/workflows/ci.yml
```diff
@@ -14,6 +14,12 @@ jobs:
     steps:
       - name: Checkout cedar-java
         uses: actions/checkout@v4
+      - name: Checkout cedar
+        uses: actions/checkout@v4
+        with:
+          repository: cedar-policy/cedar
+          ref: main
+          path: /home/runner/work/cedar-java/cedar-java/cedar
```
khieta, line 22: Is the hardcoded path ok here? Is it guaranteed to always be the same in CI?

andrewmwells-amazon: I doubt it changes often, so even if it eventually does I prefer this to keeping the config script around.

CedarJava/build.gradle
andrewmwells-amazon: At least personally, I prefer to set this here and then run builds in Intellij rather than exporting this variable in a terminal.

.github/workflows/ci.yml
exceptionfactory: As an alternative, a relative path could also work in this context. The full path should not change often, if at all, unless the runner image moves to a new major version.

PR #123, opened 2024-04-10
Updates for cedar-policy/#714

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/model/PartialAuthorizationResponse.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/model/PartialAuthorizationResponse.java
@@ -6,46 +6,107 @@ import com.cedarpolicy.model.AuthorizationResponse.Decision;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
 @Experimental(ExperimentalFeature.PARTIAL_EVALUATION)
 public class PartialAuthorizationResponse {
     private Decision decision;
-    private Set<String> satisfied;
-    private Set<String> errored;
-    private Set<String> may_be_determining;
-    private Set<String> must_be_determining;
-    private Map<String, JsonNode> residuals;
-    private Set<String> nontrivial_residuals;
+    private ImmutableSet<String> satisfied;
+    private ImmutableSet<String> errored;
+    private ImmutableSet<String> may_be_determining;
+    private ImmutableSet<String> must_be_determining;
+    private ImmutableMap<String, JsonNode> residuals;
+    private ImmutableSet<String> nontrivial_residuals;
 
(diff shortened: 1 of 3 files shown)


CedarJava/src/main/java/com/cedarpolicy/model/PartialAuthorizationResponse.java
```diff
@@ -3,128 +3,64 @@
...
-public abstract class PartialAuthorizationResponse {
-    private final Diagnostics diagnostics;
+public class PartialAuthorizationResponse {
+    private Decision decision;
+    private Set<String> satisfied;
+    private Set<String> errored;
+    private Set<String> may_be_determining;
+    private Set<String> must_be_determining;
+    private Map<String, JsonNode> residuals;
+    private Set<String> nontrivial_residuals;
```
cdisselkoen, line 23: Should these be `final`?  I see that `diagnostics` was previously.  (Not a Java person and don't really know when we should/shouldn't use `final`)

CedarJava/src/main/java/com/cedarpolicy/model/PartialAuthorizationResponse.java
```diff
@@ -3,128 +3,64 @@
...
+        this.nontrivial_residuals = new HashSet<String>(nontrivial_residuals);
+    }
+
+    public Decision getDecision() {
+        return this.decision;
+    }
+
+    public Set<String> getResiduals() {
+            return this.residuals.keySet();
+    }
```
cdisselkoen, line 42: Should we have public getters for `errored`, `may_be_determining`, etc?

CedarJava/src/main/java/com/cedarpolicy/model/PartialAuthorizationResponse.java
```diff
@@ -3,128 +3,64 @@
...
+        this.errored = new HashSet<String>(errored);
+        this.may_be_determining = new HashSet<String>(may_be_determining);
+        this.must_be_determining = new HashSet<String>(must_be_determining);
+        this.residuals = new HashMap<String, JsonNode>(residuals);
+        this.nontrivial_residuals = new HashSet<String>(nontrivial_residuals);
+    }
+
+    public Decision getDecision() {
+        return this.decision;
+    }
```
cdisselkoen, line 38: should this have docs, eg, noting what returning `null` means?

cdisselkoen (approved):

CedarJava/src/main/java/com/cedarpolicy/model/PartialAuthorizationResponse.java
shaobo-he-aws: Done.

CedarJava/src/main/java/com/cedarpolicy/model/PartialAuthorizationResponse.java
shaobo-he-aws: Done.

CedarJava/src/main/java/com/cedarpolicy/model/PartialAuthorizationResponse.java
andrewmwells-amazon: IIUC `final` just means it won't change after construction, which seems appropriate here.

andrewmwells-amazon (approved):

Issue #126, opened 2024-04-19
CI does not run tests in CedarJavaFFI


cdisselkoen:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar-java/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar-java/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Other

### Describe the bug

`CedarJavaFFI` has its own tests, but the CI in this repo does not run them; as a result, they are broken on `main` as of this writing.  We should either modify the CI to run these tests, or remove the tests if we don't think they add value and prefer to have only the tests in `CedarJava`.

### Expected behavior

`main` does not contain broken tests.

### Reproduction steps

1. Checkout main
2. `cd CedarJavaFFI`
3. `cargo test`

cdisselkoen:
Somewhat related and probably could be addressed at the same time: CedarJava and/or CedarJavaFFI could use more tests for error paths, ensuring that various failure cases are properly reported across the FFI boundary (both sides agree on the structure of JSON for these error cases).

khieta:
Closed by #128?

cdisselkoen:
Yes

PR #127, opened 2024-04-19
update for cedar#794

cdisselkoen:
*Issue #, if available:*

cedar-policy/cedar#794

*Description of changes:*

Also a couple unrelated nits.

CedarJavaFFI tests are currently failing due to #126.

### DIFF
--- a/CedarJavaFFI/src/answer.rs
+++ b/CedarJavaFFI/src/answer.rs
@@ -20,9 +20,8 @@ use serde::{Deserialize, Serialize};
 #[serde(tag = "success")]
 #[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
 #[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
-/// As of this writing, this is only used by `json_parse_entity_uid()`.
-/// We could remove this and just specialize the output of `json_parse_entity_uid()`.
-pub enum InterfaceResult {
+/// Generic Answer type designed for serialization.
+pub enum Answer {
     /// The call succeeded
     #[serde(rename = "true")]
     Success {
@@ -49,7 +48,7 @@ pub enum InterfaceResult {
     },
 }
 
-impl InterfaceResult {
+impl Answer {
     /// A successful result
     pub fn succeed<T: Serialize>(value: T) -> Self {
         serde_json::to_string(&value).map_or_else(
@@ -58,7 +57,7 @@ impl InterfaceResult {
         )
     }
 
-    /// An "internal failure" result; see docs on [`InterfaceResult::Failure`]
+    /// An "internal failure" result; see docs on [`Answer::Failure`]
     pub fn fail_internally(message: String) -> Self {
         Self::Failure {
             is_internal: true,
@@ -67,7 +66,7 @@ impl InterfaceResult {
     }
 
     /// A failure result that isn't internal; see docs on
-    /// `InterfaceResult::Failure`
+    /// `Answer::Failure`
     pub fn fail_bad_request(errors: Vec<String>) -> Self {
         Self::Failure {
             is_internal: false,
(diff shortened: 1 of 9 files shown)


CedarJavaFFI/src/interface_result.rs
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(tag = "success")]
+#[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
+#[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
+/// As of this writing, this is only used by `json_parse_entity_uid()`.
+/// We could remove this and just specialize the output of `json_parse_entity_uid()`.
```
khieta, line 24: Do we think we'll add more operations with this return type? If not, I'd lean towards specializing the output of `json_parse_entity_uid()`

khieta (approved):

CedarJavaFFI/src/interface_result.rs
cdisselkoen: I think we'll add more operations in general, in the future.  I am unclear whether they will/should use this return type or not

khieta: A fair question. Let's just leave it in the Java code for now. But for consistency with our other data types, I'm thinking that we just want it to be named `Result` or `Answer` (no "interface"). I'd also prefer not to have the internal/non-internal failure distinction, but I don't feel strongly.

cdisselkoen: Hmm my comment about only used by `json_parse_entity_uid()` appears to be wrong.  I suspect though that post-793 and friends, in some error cases our code is returning the wrong error structure.  Java code probably needs more tests with invalid call JSON.

cdisselkoen: I think it makes the most sense to just rename `InterfaceResult` to `Answer`, and leave the rest for followups.  I've done that in the latest revision of this PR

khieta (approved):

cdisselkoen:
I'm not sure what is causing the current CI failure.

CedarJavaFFI/src/answer.rs
```diff
@@ -0,0 +1,76 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(tag = "success")]
+#[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
```
john-h-kastner-aws, line 21: Should the Java FFI create have these `wasm` attributes?

john-h-kastner-aws:
> I'm not sure what is causing the current CI failure.

It looks like `Build on Linux` is trying to run `/home/runner/.cache/cargo-zigbuild/0.18.3/zigcc-aarch64-apple-darwin.sh`. Could be a problem.

cdisselkoen:
That's super weird because you wouldn't think any of the changes in this PR could be affecting the zig build (I think?)

john-h-kastner-aws:
Zig did release a new version 4 days ago: https://ziglang.org/download/0.12.0/release-notes.html

john-h-kastner-aws:
Confirmed build works on Linux with zig 0.11.0 from https://ziglang.org/download/

cdisselkoen:
Alright, build has passed now, after I both downgraded to zig 0.11.0 and also locked the version of cargo-zigbuild to the current latest for good measure.

john-h-kastner-aws (approved):
It works!

PR #133, opened 2024-05-01
update for cedar#800

cdisselkoen:
*Issue #, if available:*

cedar-policy/cedar#800

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/model/ValidationResponse.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/model/ValidationResponse.java
@@ -21,7 +21,6 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableList;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


CedarJava/src/main/java/com/cedarpolicy/model/DetailedError.java
```diff
@@ -0,0 +1,93 @@
+package com.cedarpolicy.model;
```
aaronjeline, line 1: Do we want to drop an explicit note that this is an encoding of `miette::Report`?

aaronjeline (approved):

CedarJava/src/main/java/com/cedarpolicy/model/DetailedError.java
cdisselkoen: I think that's a Rust implementation detail.  As far as the Java is concerned, this is just the structure we get over JSON from Rust.  Having that comment on the Rust side is good though

CedarJava/src/main/java/com/cedarpolicy/model/AuthorizationSuccessResponse.java
```diff
@@ -95,6 +95,29 @@ public AuthorizationSuccessResponse(
         this.diagnostics = diagnostics;
     }
 
+    /** Error (or warning) which occurred in a particular policy during authorization */
```
khieta, line 98: Nit: Currently there are no warnings returned at authorization time. Warnings would instead be captured in the `AuthorizationResponse` struct.

CedarJava/src/main/java/com/cedarpolicy/model/ValidationResponse.java
```diff
@@ -28,62 +28,87 @@
 
 /** Result of a validation request. */
 public final class ValidationResponse {
-    /** Exactly one of `results` or `errors` should be None. */
+    /** Is this a success or a failure response */
+    @JsonProperty("type")
+    public final SuccessOrFailure type;
+    /** This will be present if and only if `type` is `Success`. */
     public final Optional<ValidationResults> results;
```
khieta, line 35: Nit: To match the naming in AuthorizationReponse.java
```suggestion
    public final Optional<ValidationSuccessResponse> success;
```

khieta (approved):
Left some non-blocking comments.

CedarJava/src/main/java/com/cedarpolicy/model/ValidationResponse.java
cdisselkoen: Renamed and pushed

khieta (approved):

PR #142, opened 2024-05-09
Add support for new schema syntax

### LINKED ISSUE #130: Add support for the new schema syntax
### Category

User level API changes

### Describe the feature you'd like to request

Add support for new schema syntax introduced in Cedar v3.1. This is required to support the new integration tests in https://github.com/cedar-policy/cedar-integration-tests.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

khieta:
*Issue #, if available:*

Resolves #130 

*Description of changes:*

* Added support for the new schema syntax
* Moved the tests from `interface.rs` to a separate `tests.rs` file
* Removed some unused code from the Rust interface
* Updated integration testing code to use the tests in https://github.com/cedar-policy/cedar-integration-tests


CedarJava/src/main/java/com/cedarpolicy/AuthorizationEngine.java
```diff
@@ -81,6 +81,6 @@ public interface AuthorizationEngine {
      * @return The Cedar language major version supported
      */
     static String getCedarLangVersion() {
-        return "3.0";
+        return "3.1";
```
cdisselkoen, line 84: probably 3.2 or 4.0?  This depends on FFI changes not available in 3.1 right?

CedarJava/src/main/java/com/cedarpolicy/model/schema/Schema.java
```diff
@@ -21,86 +21,127 @@
...
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import java.io.IOException;
-import java.util.Objects;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import java.util.Optional;
 
 /** Represents a schema. */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
```
cdisselkoen, line 30: what does this do?

CedarJava/src/main/java/com/cedarpolicy/model/schema/Schema.java
```diff
@@ -21,86 +21,127 @@
...
+        /**
+         * Cedar schema format. See <a href=
+         * "https://docs.cedarpolicy.com/schema/human-readable-schema.html">https://docs.cedarpolicy.com/schema/human-readable-schema.html</a>
+         */
+        Json,
+        /**
+         * Cedar JSON schema format. See <a href=
+         * "https://docs.cedarpolicy.com/schema/json-schema.html">https://docs.cedarpolicy.com/schema/json-schema.html</a>
+         */
+        Human,
```
cdisselkoen, line 141: Doc comments seem to be switched

CedarJavaFFI/src/interface.rs
```diff
@@ -147,45 +142,25 @@ fn jni_failed(env: &mut JNIEnv<'_>, e: &dyn Error) -> jvalue {
...
+/// public string-based JSON interface to parse a schema
+#[jni_fn("com.cedarpolicy.model.schema.Schema")]
+pub fn parseJsonSchemaJni<'a>(mut env: JNIEnv<'a>, _: JClass, schema_jstr: JString<'a>) -> jvalue {
+    match parse_json_schema_internal(&mut env, schema_jstr) {
+        Ok(v) => v.as_jni(),
+        Err(e) => jni_failed(&mut env, e.as_ref()),
     }
 }
 
 /// public string-based JSON interface to parse a schema
```
cdisselkoen, line 154: JSON interface, for the non-JSON schema? Might be worth clarifying

CedarJavaFFI/src/interface.rs
```diff
@@ -201,6 +176,22 @@ fn parse_schema_internal<'a>(
...
+    schema_jstr: JString<'a>,
+) -> Result<JValueOwned<'a>> {
+    if schema_jstr.is_null() {
+        raise_npe(env)
+    } else {
+        let schema_jstring = env.get_string(&schema_jstr)?;
+        let schema_string = String::from(schema_jstring);
+        match Schema::from_str_natural(&schema_string) {
+            Err(e) => Err(Box::new(e)),
+            Ok(_) => Ok(JValueGen::Object(env.new_string("Success")?.into())),
```
cdisselkoen, line 190: Returns the string `Success` on success, but on the Java side, `Schema.parse()` seems to expect it to return the schema text?

khieta:
It turned out to not be too difficult to add support for the new integration tests, so I added that change to this PR.

CedarJava/src/main/java/com/cedarpolicy/model/schema/Schema.java
khieta: So that if `schemaJson` is empty, then instead of serializing it as `"json": null` (which makes the Rust parsers angry), Jackson ignores it.

CedarJava/src/main/java/com/cedarpolicy/AuthorizationEngine.java
khieta: I'll update to 4.0 since I'm expecting these changes to be released then.

CedarJava/src/main/java/com/cedarpolicy/model/schema/Schema.java
khieta: Oops

CedarJavaFFI/src/interface.rs
khieta: Ah, that is indeed not doing the right thing (and testing didn't catch it since we don't have any tests that do something with a parsed schema). Will fix in next commit.

cdisselkoen (approved):

CedarJavaFFI/src/interface.rs
cdisselkoen: Curious why return the schema contents?  Java already had a string containing the schema contents, I think?

khieta: Yes, Java already has the string -- the only new information "returned" by the parsing function is whether an exception was thrown or not. Looking at it again, it likely saves some time not to pass back the string, so I'll revert to that.

CedarJava/src/test/java/com/cedarpolicy/SharedIntegrationTests.java
```diff
@@ -187,19 +190,20 @@ private static class JsonEntity {
...
-       "tests/example_use_cases_doc/5b.json",
+       "tests/decimal/1.json",
+       "tests/decimal/2.json",
+       "tests/example_use_cases/1a.json",
+       "tests/example_use_cases/2a.json",
+       "tests/example_use_cases/2b.json",
+       "tests/example_use_cases/2c.json",
+       "tests/example_use_cases/3a.json",
+       "tests/example_use_cases/3b.json",
+       "tests/example_use_cases/3c.json",
```
andrewmwells-amazon, line 201: [nit] Can we keep the comment explaining why 4c isn't run?

andrewmwells-amazon (approved):

CedarJava/src/test/java/com/cedarpolicy/SharedIntegrationTests.java
khieta: I deleted it because that test was not moved over to https://github.com/cedar-policy/cedar-integration-tests -- do you think we should maintain this test in our integration test repo even if it's not used?

PR #151, opened 2024-05-29
update for cedar#921

cdisselkoen:
*Issue #, if available:*

cedar-policy/cedar#921

### DIFF
--- a/CedarJavaFFI/src/objects.rs
+++ b/CedarJavaFFI/src/objects.rs
@@ -235,7 +235,7 @@ impl<'a> JEntityId<'a> {
 
     /// Decode the object into its string representation
     pub fn get_string_repr(&self) -> String {
-        self.id.to_string()
+        self.id.as_ref().to_string()
     }
 }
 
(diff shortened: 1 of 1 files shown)


shaobo-he-aws (approved):

khieta (approved):

CedarJavaFFI/src/objects.rs
```diff
@@ -235,7 +235,7 @@ impl<'a> JEntityId<'a> {
 
     /// Decode the object into its string representation
     pub fn get_string_repr(&self) -> String {
-        self.id.to_string()
+        self.id.as_ref().to_string()
```
khieta, line 238: Why not `.escaped()`? That would have been the original behavior, right?

cdisselkoen: True, and I can change it.  I thought that perhaps folks manipulating the string in Java might expect to manipulate the unescaped string.  But I suppose no one has complained about the current behavior

khieta: No strong preference, I was just wondering if the behavior change was intentional. I think both options are reasonable, but might be better stick with the status quo 🤷‍♀️

cdisselkoen: Changed to `.escaped()`

shaobo-he-aws (approved):

PR #165, opened 2024-06-27
Adding PolicySet parsePolicies method

mark-creamer-amazon:
https://github.com/cedar-policy/cedar-java/issues/38

## CedarJava
* Added static `PolicySet parsePolicies(Path filePath)` and `PolicySet parsePolicies(String policiesString)` methods to `PolicySet`.
* Added `PolicySetTests` file that tests the File path and string methods and some exception cases.
* Adding test `cedar` policy files in `src/test/resources` directory to use for testing.
## CedarJavaFFI
* Added JNI method `parsePoliciesJni` that takes a Java String for policies and templates, uses [`PolicySet::from_str`](https://docs.rs/cedar-policy/latest/src/cedar_policy/api.rs.html#2383-2416) to parse the string into `policies` and `templates` and then constructs a Java `PolicySet` with those parameters. Template instantiations are omitted.


CedarJava/src/main/java/com/cedarpolicy/model/slice/PolicySet.java
```diff
@@ -49,4 +60,32 @@ public PolicySet(Set<Policy> policies, Set<Policy> templates, List<TemplateInsta
         this.templates = templates;
         this.templateInstantiations = templateInstantiations;
     }
+
+    /**
+     * Parse a multiple policies and templates from a file into a PolicySet.
```
khieta, line 65:
```suggestion
     * Parse multiple policies and templates from a file into a PolicySet.
```

CedarJava/src/test/resources/duplicate_policies.cedar
```diff
@@ -0,0 +1,13 @@
+@id("Policy")
```
khieta, line 1: I don't see this test case used above, but IIRC that this _will not_ be an error. The `@id()` annotation is given special meaning in the CLI, but not the Cedar library itself. So this annotation isn't actually setting the policy id here. The two policies are assigned default distinct ids, like "policy0" and "policy1".

khieta (commented):
Could you also update the integration tests to use these new APIs instead of hacky splitting on semicolons? Relevant code is here: https://github.com/cedar-policy/cedar-java/blob/8baec5e7b6c98affde9b6ca94a55c86c6fda79f4/CedarJava/src/test/java/com/cedarpolicy/SharedIntegrationTests.java#L284

CedarJava/src/test/resources/duplicate_policies.cedar
mark-creamer-amazon: Ah I see. I had initially used this to test that the Rust parsing would output a set with a single member since the policy content was identical, but I see that that's not the case, and multiple policies with different IDs but the same source are returned. I removed the tests, but forgot to remove this file as well.

shaobo-he-aws (approved):

khieta (approved):
Lgtm

PR #172, opened 2024-07-03
Adding Policy toJson method

mark-creamer-amazon:
https://github.com/cedar-policy/cedar-java/issues/171

## Policy.java
* Adding `toJson` method that calls a new JNI method `toJsonJni` to return the JSON String for the Policy.

## interface.rs
* Added JNI method `toJsonJni` that takes policy source, constructs a rust Policy struct and returns the Json representation.


khieta (approved):
Looks fine. Approving for now, although I'd like to see the test cases added.

As discussed offline: I think it's fine to merge the current approach, even though we'll (likely) be adding support for this in the Json FFI in the future. Once the FFI is updated, we can file an issue to update this code as well.

shaobo-he-aws (approved):

PR #180, opened 2024-07-17
Add support for cedar-policy-formatter

felixzheng98:
*Issue #, if available:*

*Description of changes:*

This PR adds support for cedar-policy-formatter. 

 * Add dependency to `cedar-policy-formatter` on the FFI side
 * Add jni for `cedar_policy_formatter::policies_str_to_pretty`
 * Create `PolicyFormatter` on Java side


CedarJava/src/test/java/com/cedarpolicy/PolicyFormatterTests.java
```diff
@@ -0,0 +1,50 @@
...
+import com.cedarpolicy.model.exception.InternalException;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class PolicyFormatterTests {
+
+  @Test
+  public void testPoliciesStrToPretty() throws Exception {
```
shaobo-he-aws, line 13: Nit: I'd prefer that test policies (original and formatted) are stored in files, which makes it much easier to update them.

felixzheng98: Sounds good, I've updated the tests to use the resource files.

mark-creamer-amazon (approved):

CedarJava/src/main/java/com/cedarpolicy/formatter/PolicyFormatter.java
```diff
@@ -0,0 +1,17 @@
...
+public final class PolicyFormatter {
+
+  static {
+    LibraryLoader.loadLibrary();
+  }
+
+  private PolicyFormatter() {
+  }
+
+  public static native String policiesStrToPretty(String policies)
```
shaobo-he-aws, line 15: More nit: I wonder if it makes sense to make it a method of `PolicySet`. In Rust, we cannot do it because the formatter and `PolicySet` are in different packages. But we should be able to do it here.

shaobo-he-aws (approved):
LGTM with minor nits.

CedarJava/src/main/java/com/cedarpolicy/formatter/PolicyFormatter.java
felixzheng98: Do you mean to move this method into the `PolicySet` class?

I don't think that is best because this method is strictly for formatting. The output is a String.
This makes it not as directly related to `PolicySet`, so I think it should be belong in this separate class.

felixzheng98:
Please merge this for me when convenient as I don't have permissions to merge. Thank you!

shaobo-he-aws:
> Please merge this for me when convenient as I don't have permissions to merge. Thank you!

Merged. Thank you for your contribution.

CedarJava/src/main/java/com/cedarpolicy/formatter/PolicyFormatter.java
shaobo-he-aws: I'm fine with it as well.

PR #183, opened 2024-08-05
Changes needed by cedar-policy/cedar#1114

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*


john-h-kastner-aws (approved):
Are there changes we should make to the Java API?

shaobo-he-aws:
> Are there changes we should make to the Java API?

Good catch. Yes. I'll update them as well.

andrewmwells-amazon (approved):

PR #193, opened 2024-08-29
Remove test that shouldn't pass

andrewmwells-amazon:
*Issue #, if available:*

*Description of changes:*
Remove test case that shouldn't pass


shaobo-he-aws (approved):

khieta (approved):
This test will still pass in the version though -- we didn't fix the template constructor until v3.3.

andrewmwells-amazon:
> This test will still pass in the version though -- we didn't fix the template constructor until v3.3.

It's failing on the internal CI

PR #202, opened 2024-09-11
Add support for cedar entities validation.

amzn-mdamine:
Issue #, if available:

Description of changes:

This PR adds support for cedar-entity-validation.

Refactor Entity json serialization similar to release/4.0.x
- Add entity validation in rust
- Add entity validation in java
- Add unit test in java and rust.

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/AuthorizationEngine.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/AuthorizationEngine.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2022-2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -74,6 +74,17 @@ public interface AuthorizationEngine {
      * @throws AuthException if any internal errors occurred while validating the policies.
      */
     ValidationResponse validate(ValidationRequest request) throws AuthException;
+
+    /**
+     * Asks whether the entities in the given {@link EntityValidationRequest} <code>q</code> are correct
+     * when validated against the schema it describes.
+     *
+     * @param request The request containing the entities to validate and the schema to validate them
+     *                against.
+     * @throws BadRequestException if any errors were found in the syntax of the entities.
+     * @throws AuthException if any internal errors occurred while validating the policies.
+     */
+    void validateEntities(EntityValidationRequest request) throws AuthException;
 
     /**
      * Get the Cedar language major version (e.g., "1.2") used by this CedarJava library.
--- a/CedarJava/src/test/java/com/cedarpolicy/EntityValidationTests.java
+++ b/CedarJava/src/test/java/com/cedarpolicy/EntityValidationTests.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2022-2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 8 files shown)


andrewmwells-amazon (approved):

shaobo-he-aws (approved):

CedarJava/src/main/java/com/cedarpolicy/AuthorizationEngine.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2022-2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andrewmwells-amazon, line 2: We can change all of these to just "Copyright Cedar Contributors"

amzn-mdamine: Yes, I saw that in release/4.0.0. I think it is better than keeping track of these dates. However, I believe the current branch uses that copyright format so I wanted to keep things consistant

CedarJava/src/test/java/com/cedarpolicy/EntityValidationTests.java
```diff
@@ -0,0 +1,118 @@
...
+    @Test
+    public void testEntityWithUnknownAttribute() throws AuthException {
+        Entity entity = EntityValidationTests.entityGen.arbitraryEntity();
+        entity.attrs.put("test", new PrimBool(true));
+
+        EntityValidationRequest request = new EntityValidationRequest(
+                ROLE_SCHEMA, List.of(entity));
+
+
+        AuthException exception = assertThrows(AuthException.class, () -> engine.validateEntities(request));
```
amzn-mdamine, line 76: I was wondering if this is the exception that should be thrown for this test.

CedarJava/src/test/java/com/cedarpolicy/EntityValidationTests.java
```diff
@@ -0,0 +1,118 @@
...
+        // Create a cyclic parent relationship between the entities
+        childEntity.parentsEUIDs.add(parentEntity.getEUID());
+        parentEntity.parentsEUIDs.add(childEntity.getEUID());
+
+        EntityValidationRequest request = new EntityValidationRequest(
+                ROLE_SCHEMA,
+                List.of(parentEntity, childEntity));
+
+        // Act & Assert
+        AuthException exception = assertThrows(AuthException.class, () -> engine.validateEntities(request));
```
amzn-mdamine, line 101: Same question as above

andrewmwells-amazon: `InternalException` seems like the best fit.

CedarJava/src/test/java/com/cedarpolicy/EntityValidationTests.java
andrewmwells-amazon: `InternalException` seems like the best fit.

CedarJavaFFI/src/interface.rs
```diff
@@ -119,12 +121,42 @@ fn call_cedar(call: &str, input: &str) -> String {
...
+    schema: Value,
+    #[serde(rename = "entities")]
+    entities: Value,
+}
+
+/// public string-based JSON interface to be invoked by FFIs. Takes in a `ParseEUIDCall`, parses it and (if successful)
+/// returns a serialized `ParseEUIDOutput`
+pub fn json_validate_entities(input: &str) -> InterfaceResult {
+    match from_str::<ValidateEntityCall>(&input).map_err(|_| "Failed to parse JSON") {
+        Err(e) => InterfaceResult::fail_internally(format!("error parsing call object: {e:}")),
```
andrewmwells-amazon, line 143: I think this should be a `fail_bad_request`

CedarJava/src/test/java/com/cedarpolicy/EntityValidationTests.java
khieta: `BadRequestException` is fine, or you could make a custom exception, along the lines of `InvalidEUIDException`.

CedarJava/src/test/java/com/cedarpolicy/EntityValidationTests.java
khieta: Same comment as above

khieta (approved):
Changes look good 🙂 

Remember to port this to `release/3.4.x`, `release/4.0.x`, and `main` as well. The first branch should be an easy cherry-pick. The latter two will require a little more effort since  we've made changes to the code that this PR touches. I'd recommend putting up another PR for review against `release/4.0.x`.

PR #206, opened 2024-09-12
Add support for cedar entities validation.

amzn-mdamine:
Issue #, if available:

Description of changes:

This PR adds support for cedar-entity-validation.

- Add entity validation in rust
- Add entity validation in java
- Add unit test in java and rust.

Similar to this: https://github.com/cedar-policy/cedar-java/pull/202

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/BasicAuthorizationEngine.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/BasicAuthorizationEngine.java
@@ -24,15 +24,20 @@ import java.io.IOException;
 import com.cedarpolicy.loader.LibraryLoader;
 import com.cedarpolicy.model.*;
 import com.cedarpolicy.model.exception.AuthException;
+import com.cedarpolicy.model.exception.BadRequestException;
 import com.cedarpolicy.model.exception.InternalException;
 import com.cedarpolicy.model.exception.MissingExperimentalFeatureException;
 import com.cedarpolicy.model.entity.Entity;
 import com.cedarpolicy.model.policy.PolicySet;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
+import java.util.List;
 import java.util.Set;
 
 /** An authorization engine that is compiled in process. Communicated with via JNI. */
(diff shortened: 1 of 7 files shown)


CedarJavaFFI/src/interface.rs
```diff
@@ -125,6 +129,43 @@ pub(crate) fn call_cedar(call: &str, input: &str) -> String {
...
+        Ok(schema) => {
+            match Entities::from_json_value(validate_entity_call.entities, Some(&schema)) {
+                Err(error) => {
+                    let err_message = match error {
+                        EntitiesError::Serialization(err) => err.to_string(),
+                        EntitiesError::Deserialization(err) => err.to_string(),
+                        EntitiesError::Duplicate(err) => err.to_string(),
+                        EntitiesError::TransitiveClosureError(err) => err.to_string(),
+                        EntitiesError::InvalidEntity(err) => err.to_string(),
+                    };
```
khieta, line 158: Does `error.to_string()` not work?

CedarJava/src/main/java/com/cedarpolicy/BasicAuthorizationEngine.java
```diff
@@ -100,6 +117,40 @@ private static <REQ, RESP> RESP call(String operation, Class<RESP> responseClass
         }
     }
 
+    /**
+     * The result of processing an EntityValidationRequest.
+     */
+    @JsonIgnoreProperties({"result"})  // Ignore only the 'result' field
+    private static final class EntityValidationResponse {
```
khieta, line 124: We usually put these types of definitions in a separate file (like you did for `EntityValidationRequest`)

khieta (approved):
Looks good overall. Just left some minor comments.

CedarJavaFFI/src/interface.rs
amzn-mdamine: It  gives a generic message rather than detailed one.

CedarJava/src/main/java/com/cedarpolicy/BasicAuthorizationEngine.java
amzn-mdamine: I can extract this class, but I’m wondering whether it should be public or package-private. Should the package clients have access to this class? Currently, when clients use the entityValidation function, it returns void and throws a BadRequestException if the entities are not valid.

What do you suggest?

khieta: Seems like it could be package-private since the structure is just to aid deserialization.

CedarJavaFFI/src/interface.rs
khieta: Ah, this is a case where the useful information is hidden inside the `help` text... let me find the right incantation to use to get things to print properly

amzn-mdamine: Understood, I will try it.

CedarJava/src/main/java/com/cedarpolicy/BasicAuthorizationEngine.java
amzn-mdamine: Ok I will implement the necessary changes.

shaobo-he-aws (approved):
LGTM. Will re-approve it once the requested changes are implemented.

CedarJavaFFI/src/interface.rs
amzn-mdamine: I tried this and it seems that help msg is not provided by all of them. help return option and it is not set for some of them.

PR #227, opened 2024-10-01
JDK8 compatibility for non-test classes

cdisselkoen:
Currently it's not that painful to maintain compatibility with JDK8 in non-test classes.


mark-creamer-amazon (approved):
These changes look good 👍

shaobo-he-aws (approved):

PR #238, opened 2024-10-15
Adding support for entity tags

mark-creamer-amazon:
### Use Case
Entity tagging is new in Cedar 4.2.x, but CedarJava 4.2.x is not able to leverage it with the current Entity object and serializer.

### Details

* Adding `tags` field to `Entity.java` to support entities with tags
* Updated `EntitySerializer.java` to make sure the tags field makes it across the JNI layer and into the AuthorizationRequest string we pass.
* Quality of life improvement: returning the error message in the case of Validation errors in the corpus/json tests in `executeJsonValidationTest`

### Testing

* Added `testEntityWithUnknownTag` to  `EntityValidationTests.java`
* Created a temporary extra json test among `JSON_TEST_FILES` that tested e2e allowing an action contingent on a tag being present in the resource.

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/model/entity/Entity.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/model/entity/Entity.java
@@ -25,8 +25,8 @@ import java.util.stream.Collectors;
 /**
  * An entity is the kind of object about which authorization decisions are made; principals,
  * actions, and resources are all a kind of entity. Each entity is defined by its entity type, a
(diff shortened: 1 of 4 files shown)


cdisselkoen (approved):

shaobo-he-aws (approved):

cdisselkoen:
Can we make sure to make the same change on `main` as well?

mark-creamer-amazon:
> Can we make sure to make the same change on `main` as well?

Created here: https://github.com/cedar-policy/cedar-java/pull/239

PR #256, opened 2024-11-14
Fix isAuthorizedPartial due to rust interface changes

amzn-mdamine:
There was a commit that changes the rust interface for isAuthorizedPartial method. The current PR fix the problem by making the necessary changes.

 isAuthorizedPartial was never working since the rust change was made so it is safe to assume that no client use it for now.

There were unit tests for isAuthorizedPartial method however they were not being executed correctly.


cdisselkoen (approved):
Great catch, thanks for this.

CedarJava/src/test/java/com/cedarpolicy/AuthTests.java
```diff
@@ -92,21 +92,21 @@ public void residual() {
         assumePartialEvaluation(() -> {
             try {
                 final PartialAuthorizationResponse response = auth.isAuthorizedPartial(q, policySet, new HashSet<>());
-                assertTrue(response.getDecision() == null);
-                assertEquals("p0", response.getResiduals().entrySet().iterator().next().getKey());
+                assertTrue(response.success.orElseThrow().getDecision() == null);
```
adpaco-aws, line 95: How can the decision be `null` when the response is a success?

adpaco-aws (approved):
Thanks!

Marking this "ready to review" since it's still a draft PR. Or are you still working on it?

CedarJava/src/test/java/com/cedarpolicy/AuthTests.java
amzn-mdamine: This test already existed; I haven't changed its assertion, I only fixed the result.

As for the decision, in Cedar partial authorization, when there are unknowns (in this case, the request has the principal as unknown), a decision is not able to be made. Thus, the decision is null.

amzn-mdamine:
I think the PR is ready to be merged if there are no concerns

PR #265, opened 2024-12-17
Add support for building a Linux / musl native library

hildo:
*Issue # 264*

Include building a native library for linux that uses musl ABI


hildo:
Weird.  Looks like the aarch64 is failing... that was a nice to have, so I will pull it out and hope that the x86 will link correctly.

hildo:
Hmm, it doesn't look like just adding a new condition will work.  So this may not be as simple as I had hoped.  Will likely close this out and try alternative approaches.  Apologies for the interruption

hildo:
Closing this request.  I don't think muy understanding is good enough, where what I have proposed will work.  Will find another way around my problem.

PR #273, opened 2025-01-02
Update Rust to 1.83

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/CedarJava/build.gradle
+++ b/CedarJava/build.gradle
@@ -120,7 +120,7 @@ tasks.register('installCargoZigbuild', Exec) {
     group 'Build'
     description 'Installs Cargo Zigbuild for Rust compilation.'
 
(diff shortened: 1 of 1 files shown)


cdisselkoen (commented):
Do we want this in our `build.gradle`?  Theoretically a new Rust release could then break the build of a released version of CedarJava?  Is it better to pin to a known-working Rust version (here, 1.81)?

shaobo-he-aws:
> Do we want this in our `build.gradle`? Theoretically a new Rust release could then break the build of a released version of CedarJava? Is it better to pin to a known-working Rust version (here, 1.81)?

Yeah. It seems 1.83 (current stable) still doesn't work. I will close this PR if 1.82 doesn't work as well.

cdisselkoen (approved):

john-h-kastner-aws (approved):

PR #275, opened 2025-01-09
Exposes policy effect for static policies and templates

muditchaudhary:
## Description of Changes
This PR adds functionality to expose the `effect` of Cedar policies, aligning with the capabilities available in Cedar Rust's [`Policy`](https://docs.rs/cedar-policy/latest/cedar_policy/struct.Policy.html#method.effect) and [`Template`](https://docs.rs/cedar-policy/latest/cedar_policy/struct.Template.html#method.effect) implementations.

## Changes
- Implements `effect()` method to determine if a policy is a `"permit"` or `"forbid"` policy
- Exposes `effect` for both static policies and templates
- Adds unit tests

## Implementation Note
Cedar Java currently uses a unified `Policy` class for both static policies and templates, unlike Cedar Rust which separates these concerns into distinct `Policy` and `Template` structs. To maintain compatibility while exposing this functionality:

1. The `effect()` method first attempts to parse the input as a static policy and gets the effect
2. If getting the effect as static policy fails, it attempts to parse as a template
3. Returns the appropriate effect based on the successful parse

## Future Considerations
Refactor Cedar Java to mirror Cedar Rust's separation of `Policy` and `Template` into distinct classes, allowing for better handling of each policy type.

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
@@ -79,6 +79,24 @@ public class Policy {
     }
 
     /**
+     * Returns the effect of a policy. 
+     * 
+     * Determines the policy effect by attempting static policy first, then template. 
+     * In future, it will only support static policies once new class is introduced for Template.
+     * 
+     * @return The effect of the policy, either "permit" or "forbid"
+     * @throws InternalException
+     * @throws NullPointerException
+     */
+    public String effect() throws InternalException, NullPointerException {
+        try {
+            return policyEffectJni(policySrc); // Get effect for static policy
+        } catch (InternalException e) {
+            return templateEffectJni(policySrc); // Get effect for template
+        }
+    }
+
+    /**
      * Get the JSON representation of the policy. Currently only supports static policies.
      */
     public String toJson() throws InternalException, NullPointerException {
@@ -106,4 +124,6 @@ public class Policy {
 
     private native String toJsonJni(String policyStr) throws InternalException, NullPointerException;
     private static native String fromJsonJni(String policyJsonStr) throws InternalException, NullPointerException;
+    private native String policyEffectJni(String policyStr) throws InternalException, NullPointerException;
+    private native String templateEffectJni(String policyStr) throws InternalException, NullPointerException;
 }
--- a/CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
@@ -20,6 +20,7 @@ import com.cedarpolicy.loader.LibraryLoader;
 import com.cedarpolicy.model.exception.InternalException;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import com.cedarpolicy.model.Effect;
 
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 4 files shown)


CedarJava/src/test/java/com/cedarpolicy/PolicyTests.java
```diff
@@ -109,4 +109,36 @@ public void policyFromJsonTest() throws InternalException {
...
+        });
+
+        // For effects not in {permit, forbid}
+        assertThrows(InternalException.class, () -> {
+            Policy p = new Policy("perm(principal == ?principal, action, resource in ?resource);", null);
+            p.toJson();
+        });
+
+        String actualPermitValue = "\"permit\"";
+        String actualForbidValue = "\"forbid\"";
```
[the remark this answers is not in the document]
muditchaudhary: I agree. This will be fixed

CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
```diff
@@ -78,6 +78,24 @@ public String toString() {
...
+     * Returns the effect of a policy. 
+     * 
+     * Determines the policy effect by attempting static policy first, then template. 
+     * In future, it will only support static policies once new class is introduced for Template.
+     * 
+     * @return The effect of the policy, either "permit" or "forbid"
+     * @throws InternalException
+     * @throws NullPointerException
+     */
+    public String effect() throws InternalException, NullPointerException {
```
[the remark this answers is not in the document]
muditchaudhary: I will create an Enum similar to the [Effect Enum in Cedar Rust](https://docs.rs/cedar-policy/latest/cedar_policy/enum.Effect.html).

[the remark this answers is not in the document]
mark-creamer-amazon: Ah absolutely, good catch

CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
```diff
@@ -78,6 +78,24 @@ public String toString() {
...
+        try {
+            return policyEffectJni(policySrc); // Get effect for static policy
```
mark-creamer-amazon: One thought I had here: does the order in which we try here matter?

I.e., we'd never see the exception thrown from `policyEffectJni`, only that from `templateEffectJni`. The policy exception would be swallowed. Is that fine?

As it stands now, I think the exceptions are similar and not specific to either static policies or templates, so we should be fine.

But say that I thought I had a `Policy` Java object for my static policy, and there was another issue with the policy with regards to this method. I would never see the relevant exception because we'd see an exception from failing to get the effect as a template.

Since we have to throw one of the exceptions anyway, there's not much we can do about it I don't think. It's an unfortunate symptom of the current "single class representing both Policy and Template" issue.

john-h-kastner-aws: IIRC, it should be sufficient to only call the `templateEffectJni` without bothering to try parsing as a policy first. This isn't any better for the perspective a caller who thinks they have a policy rather than a template, but it would avoid needing to swallow the exception here

CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
```diff
@@ -79,20 +80,20 @@ public String toString() {
...
+     *
      * @return The effect of the policy, either "permit" or "forbid"
      * @throws InternalException
      * @throws NullPointerException
      */
-    public String effect() throws InternalException, NullPointerException {
+    public Effect effect() throws InternalException, NullPointerException {
         try {
-            return policyEffectJni(policySrc); // Get effect for static policy
+            return Effect.fromString(policyEffectJni(policySrc)); // Get effect for static policy
```
mark-creamer-amazon, line 94: I'm fine with this as is, but any chance the return value of the jni method could be the enum instead of a String?

CedarJavaFFI/src/interface.rs
```diff
@@ -387,6 +387,58 @@ fn to_json_internal<'a>(env: &mut JNIEnv<'a>, policy_jstr: JString<'a>) -> Resul
...
+    policy_jstr: JString<'a>,
+) -> Result<JValueOwned<'a>> {
+    if policy_jstr.is_null() {
+        raise_npe(env)
+    } else {
+        let policy_jstring = env.get_string(&policy_jstr)?;
+        let policy_string = String::from(policy_jstring);
+        let policy = Template::from_str(&policy_string)?;
+        let policy_effect = serde_json::to_string(&policy.effect())?;
+        let policy_effect_unquoted: &str = serde_json::from_str(&policy_effect)?;
```
john-h-kastner-aws, line 435: You can probably do `policy.effect().to_string()` here

CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
mark-creamer-amazon: > I agree. One way to solve it could be to create a new exception class specifically for "Policy type and method mismatch." But as we plan to separate Policy and Template, I wonder if it would be worthwhile to create the new exception as it will probably not be thrown once the classes are separated.

Yeah agreed I don't think it warrants it either.

mark-creamer-amazon (approved):

shaobo-he-aws (approved):

PR #286, opened 2025-01-29
Creates Backward Compatible Context Class

muditchaudhary:
# Description
1. Currently, CedarJava handles Context as `Map<String,Value>` without a dedicated class. This PR creates Java equivalent class for [Context](https://docs.rs/cedar-policy/latest/cedar_policy/struct.Context.html).
2. Implements methods to create and merge Contexts within CedarJava. Currently, does not implement from_* methods.
3. Uses method overloading to make it backward compatible in authorization requests.
4. Implements tests for Context. 
5. Implements tests to verify backward compatibility in AuthorizationRequest and PartialAuthorizationRequest.

_Future Consideration_:
- Create `.copy()` for all classes that implement `Value` interface. This will allow us to create deep-copy of any object that use `Value`. Currently, we can only create shallow copies for objects such as `Context`. For example, `.getContext()` creates and returns a shallow copy of the `Map<String, Value> context` but the `Value` object can potentially be modified.

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/model/Context.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/model/Context.java
@@ -18,10 +18,11 @@ package com.cedarpolicy.model;
 
 import java.util.HashMap;
 import java.util.Collections;
-import java.util.Collections.singletonMap;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 import java.util.Map;
-import java.util.List;
 import com.cedarpolicy.value.Value;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 
 public class Context {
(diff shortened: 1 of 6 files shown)


CedarJava/src/test/java/com/cedarpolicy/AuthTests.java
```diff
@@ -52,6 +54,7 @@ private void assertAllowed(AuthorizationRequest q, PolicySet policySet, Set<Enti
         });
     }
 
+
```
jtbrain, line 57: nit: extra line

CedarJava/src/test/java/com/cedarpolicy/AuthTests.java
```diff
@@ -84,6 +147,53 @@ public void concrete() {
...
+        var auth = new BasicAuthorizationEngine();
+        var alice = new EntityUID(EntityTypeName.parse("User").get(), "alice");
+        var view = new EntityUID(EntityTypeName.parse("Action").get(), "view");
+        Map<String, Value> context = new HashMap<>();
+        context.put("authenticated", new PrimBool(true));
+        var q = PartialAuthorizationRequest.builder().principal(alice).action(view).resource(alice).context(context).build();
+        var policies = new HashSet<Policy>();
+        policies.add(new Policy("permit(principal == User::\"alice\",action,resource) when {context.authenticated == true};", "p0"));
+        var policySet = new PolicySet(policies);
+        assumePartialEvaluation(() -> {
```
jtbrain, line 161: nit: might be nice to get some new lines in here, its kind of hard to read this blob of text

jtbrain (approved):

CedarJava/src/main/java/com/cedarpolicy/model/Context.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /**
+     * Processes an Iterable of Map entries and adds them to the context.
+     *
+     * @param contextIterator The Iterable containing key-value pairs to add to the context
+     * @throws IllegalStateException if a duplicate key is found in the existing context or duplicate key found within the iterable
+     * @throws IllegalArgumentException if the contextIterator is null
+     */
+    private void fromIterable(Iterable<Map.Entry<String, Value>> contextIterator) throws IllegalStateException, IllegalArgumentException {
```
mark-creamer-amazon, line 123: nit: I think a better name could be given for this `void` method instead of `fromIterable`. Something more demonstrative and implies that the context instance variable is being mutated.

muditchaudhary: How about `addEntriesFromIterable` or `AddContextFromIterable`?

muditchaudhary: Refactored it to `mergeContextfromIterable`

mark-creamer-amazon (approved):

PR #287, opened 2025-01-30
Add JVM to enable testing of CedarJavaFFI interface methods

muditchaudhary:
# Description
- Currently only `fn callCedar` has direct tests in CedarJavaFFI, while other interface methods are tested indirectly via CedarJava. This leads to test coverage failures in CedarJavaFFI for some build systems
- Testing interface methods requires running a JVM instance from Rust
- This PR adds:
  - A utility method to create a single immutable JVM instance for use across parallel tests
  - An example test demonstrating usage of the shared JVM instance

### DIFF
--- a/CedarJavaFFI/src/interface.rs
+++ b/CedarJavaFFI/src/interface.rs
@@ -623,3 +623,33 @@ fn policies_str_to_pretty_internal<'a>(
         }
     }
 }
+
+#[cfg(test)]
+mod interface_tests {
+    use super::*;
+    use crate::jvm_test_utils::*;
+    use jni::JavaVM;
+    use std::sync::LazyLock;
+
+    // Static JVM to be used by all the tests. LazyLock for lazy thread-safe lazy initialization
+    static JVM: LazyLock<JavaVM> = LazyLock::new(|| create_jvm().unwrap());
+
+    mod policy_tests {
+        use super::*;
+
+        fn policy_effect_test_util(env: &mut JNIEnv, policy: &str, expected_effect: &str) {
+            let policy_string = env.new_string(policy).unwrap();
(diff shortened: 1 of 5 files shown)


CedarJavaFFI/src/interface.rs
```diff
@@ -623,3 +623,33 @@ fn policies_str_to_pretty_internal<'a>(
...
 }
+
+#[cfg(test)]
+mod interface_tests {
+    use super::*;
+    use crate::jvm_test_utils::*;
+    use jni::JavaVM;
+    use std::sync::LazyLock;
+
+    // Static JVM to be used by all the tests. LazyLock for lazy thread-safe lazy initialization
```
cdisselkoen, line 634:
```suggestion
    // Static JVM to be used by all the tests. LazyLock for thread-safe lazy initialization
```

cdisselkoen (approved):

CedarJavaFFI/src/interface.rs
```diff
@@ -623,3 +623,33 @@ fn policies_str_to_pretty_internal<'a>(
...
+    static JVM: LazyLock<JavaVM> = LazyLock::new(|| create_jvm().unwrap());
+
+    mod policy_tests {
+        use super::*;
+
+        fn policy_effect_test_util(env: &mut JNIEnv, policy: &str, expected_effect: &str) {
```
john-h-kastner-aws, line 640:
```suggestion
        #[track_caller]
        fn policy_effect_test_util(env: &mut JNIEnv, policy: &str, expected_effect: &str) {
```

Always nice to add for this sort of utility

https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute

john-h-kastner-aws (approved):

jtbrain (approved):

PR #293, opened 2025-02-12
Adds Entities with backward compatibility

muditchaudhary:
## Description

This PR introduces the `Entities` class to CedarJava, aligning it more closely with CedarRust's implementation. Key changes include:

1. **New `Entities` class**: Replacement for the current `Set<Entity>` representation for entity collections. This new class is equivalent to CedarRust's [`Entities`](https://docs.rs/cedar-policy/latest/cedar_policy/struct.Entities.html).

2. **Parse Entities from JSON**: The `Entities` class supports parsing entities from both JSON strings and files.

4. **Updated authorization methods**: 
   - Overloaded `isAuthorized()` and `isAuthorizedPartial()` methods to accept the new `Entities` type.
   - Maintained backward compatibility by retaining support for `Set<Entity>` in existing method signatures.

5. **Testing**:
   - Added unit tests for the new `Entities` class.
   - Implemented tests to verify backward compatibility of authorization calls using both `Entities` and `Set<Entity>`.
   - Introduced `org.skyscreamer:jsonassert:2.0-rc1` as a test dependency
      - Enables semantic comparison of JSON structures needed to test `Entities` using `JSONAssertEquals()`.
      - Provides foundation for future JSON-related testing needs (e.g., `.toJson()` methods).

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/AuthorizationEngine.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/AuthorizationEngine.java
@@ -21,6 +21,7 @@ import com.cedarpolicy.model.exception.AuthException;
 import com.cedarpolicy.model.exception.BadRequestException;
 import com.cedarpolicy.model.entity.Entity;
 import com.cedarpolicy.model.policy.PolicySet;
+import com.cedarpolicy.model.entity.Entities;
 
 import java.util.Set;
 
@@ -53,6 +54,21 @@ public interface AuthorizationEngine {
 
     /**
      * Asks whether the given AuthorizationRequest <code>q</code> is approved by the <code>policySet</code> and
+     * <code>entities</code> hierarchy given. Overloaded method to accept Entities object.
+     *
+     * @param request The request to evaluate
+     * @param policySet The policy set to evaluate against
+     * @param entities The entities to evaluate against
+     * @return The result of the request evaluation
+     * @throws BadRequestException if any errors were found in the syntax of the policies.
+     * @throws AuthException On failure to make the authorization request. Note that errors inside the
+     *     authorization engine are included in the <code>errors</code> field on the
+     *     AuthorizationResponse.
+     */
+    AuthorizationResponse isAuthorized(AuthorizationRequest request, PolicySet policySet, Entities entities) throws AuthException;
+
+    /**
+     * Asks whether the given AuthorizationRequest <code>q</code> is approved by the <code>policySet</code> and
      * <code>entities</code> given. If information required to answer is missing, residual policies are returned.
      *
      * @param request The request to evaluate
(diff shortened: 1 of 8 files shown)


CedarJava/src/main/java/com/cedarpolicy/model/entity/Entities.java
```diff
@@ -0,0 +1,92 @@
...
+
+import static com.cedarpolicy.CedarJson.objectReader;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.*;
```
mark-creamer-amazon, line 27: Nit: I'd prefer we avoid `*` imports and specify the particular packages.

CedarJava/src/main/java/com/cedarpolicy/model/entity/Entities.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public static Entities parse(Path filePath) throws IOException, JsonProcessingException {
+        String jsonString = Files.readString(filePath);
+        return new Entities(objectReader().forType(new TypeReference<Set<Entity>>() {
+        }).readValue(jsonString));
+    }
+
+    @Override
+    public String toString() {
+        return String.join("\n", this.entities.stream().map(Entity::toString).toList());
```
mark-creamer-amazon, line 90: Does this style match the behavior in Rust where each is printed on a new line?

CedarJava/src/main/java/com/cedarpolicy/BasicAuthorizationEngine.java
```diff
@@ -73,6 +83,16 @@ public PartialAuthorizationResponse isAuthorizedPartial(com.cedarpolicy.model.Pa
         }
     }
 
+    /**
+     * Overoaded method to accept Entities object
```
mark-creamer-amazon, line 87: Nit typo: `Overoaded` -> `Overloaded`

CedarJava/src/main/java/com/cedarpolicy/model/entity/Entities.java
muditchaudhary: No. The default behavior in Rust is to display the JSON version of Entities. `.toString()` for `Entities` continues to follow the same display format which was implemented for `Entity`

muditchaudhary: It seems like this format for displaying `Entity` was introduced in CedarJava 2.x. Maybe we can make the behavior similar to Rust in the next major version update as this will change the behavior of an existing public API

mark-creamer-amazon (approved):

CedarJava/src/main/java/com/cedarpolicy/AuthorizationEngine.java
```diff
@@ -21,6 +21,7 @@
 import com.cedarpolicy.model.exception.BadRequestException;
 import com.cedarpolicy.model.entity.Entity;
 import com.cedarpolicy.model.policy.PolicySet;
+import com.cedarpolicy.model.entity.Entities;
```
adpaco-aws, line 24: Can we please sort these out?

CedarJava/src/main/java/com/cedarpolicy/BasicAuthorizationEngine.java
```diff
@@ -27,6 +27,7 @@
 import com.cedarpolicy.model.exception.BadRequestException;
 import com.cedarpolicy.model.exception.InternalException;
 import com.cedarpolicy.model.exception.MissingExperimentalFeatureException;
+import com.cedarpolicy.model.entity.Entities;
 import com.cedarpolicy.model.entity.Entity;
 import com.cedarpolicy.model.policy.PolicySet;
```
adpaco-aws, line 32: Nit: This follows a different order than before. Let's be consistent.

adpaco-aws (approved):
Are there no plans to deprecate the methods using `Set<Entity>` in favor of these new ones? It's great to have these APIs but at the moment it's not clear why anyone would use one over the other.

jtbrain (approved):

PR #296, opened 2025-02-18
Adds annotation support for Policy

muditchaudhary:
## Description
This PR implements annotation getters for the `Policy` class, providing access to both static policy and template policy annotations.

## Key Changes
- Implements a Java HashMap in `CedarJavaFFI` for annotation management and JNI communication
- Adds annotation access methods to `Policy` class:
  - `getAnnotation(String key)`
  - `getAnnotations()`
- Implements lazy loading pattern for annotations:
  - Annotations initialize via JNI only on first getter call
  - Subsequent calls utilize cached annotations
- Adds test coverage for:
  - New FFI methods
  - `CedarJavaFFI` HashMap implementation
  - `Policy` annotation getters

## Implementation Details
1. Handling Static Policies and Templates
   - Uses try-catch approach to attempt static policy annotation retrieval first
   - Falls back to template annotation retrieval on failure (`errorMessage.contains("expected a static policy")`)
   - Current implementation relies on error message parsing (precedent in [PolicyTests](https://github.com/cedar-policy/cedar-java/blob/3421046fbb842879cb3c5244545621cf900e4a96/CedarJava/src/test/java/com/cedarpolicy/PolicyTests.java#L92)). Limited by information exposed by cedar-policy's `ParseError` (tracking in [#979](https://github.com/cedar-policy/cedar/issues/979))
   - Alternative: Completely consume `ParseError` for static policies (as in PR #275)
   - Temporary workaround until `Policy` and `Template` class separation

2. Lazy Loading Strategy
   - Loads annotations on-demand instead of in `Policy` constructor to preserve backward compatibility by avoiding addition of `InternalException` in constructor signature.

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
@@ -129,17 +129,26 @@ public class Policy {
 
     private void ensureAnnotationsLoaded() throws InternalException {
         if (annotations == null) {
-            this.annotations = getPolicyAnnotationsJni(this.policySrc);
+            try{
+                this.annotations = getPolicyAnnotationsJni(this.policySrc);
+            } catch (InternalException e) {
+                if (e.getMessage().contains("expected a static policy")) {
+                    this.annotations = getTemplateAnnotationsJni(this.policySrc);
+                }
+                else {
+                    throw e;
+                }
+            }
+            
         }
     }
 
     private static native String parsePolicyJni(String policyStr) throws InternalException, NullPointerException;
-    private static native String parsePolicyTemplateJni(String policyTemplateStr)
-            throws InternalException, NullPointerException;
-
+    private static native String parsePolicyTemplateJni(String policyTemplateStr) throws InternalException, NullPointerException;
     private native String toJsonJni(String policyStr) throws InternalException, NullPointerException;
     private static native String fromJsonJni(String policyJsonStr) throws InternalException, NullPointerException;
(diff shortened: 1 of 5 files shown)


CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
```diff
@@ -114,17 +115,69 @@ public static Policy parseStaticPolicy(String policyStr) throws InternalExceptio
...
+        return this.annotations.getOrDefault(key, null);
+    }
+
+    /**
+     * Ensures that the annotations map is loaded for this policy. If annotations haven't been loaded yet, attempts to
+     * load them first from static policy, then falls back to template if needed.
+     *
+     * @throws InternalException if there is an error loading or parsing the annotations
+     */
+    private void ensureAnnotationsLoaded() throws InternalException {
```
mark-creamer-amazon, line 153: Is there any reason we can't include this snippet of code that should be run once per instance in the constructor?

mark-creamer-amazon (approved):

CedarJava/src/main/java/com/cedarpolicy/model/policy/Policy.java
muditchaudhary: This snippet throws an `InternalException`, but the current constructor doesn't declare this exception. Moving this code into the constructor would require us to either add `throws InternalException` to the constructor signature, which would be a breaking change for existing users.

mark-creamer-amazon (approved):

john-h-kastner-aws (approved):
looks reasonable

PR #315, opened 2025-06-19
Added missing unit tests for interface.rs in CedarJavaFFI.

swabyra:
Wrote tests for the following methods in interface.rs :
parse_json_schema_internal   
parse_cedar_schema_internal
parse_policy_internal
parse_policy_template_internal
to_json_internal
template_effect_jni_internal
from_json_internal

Wrote internal unit tests to validate correct handling and null input strings. Strengthens test coverage for parsing for these previously untested methods.

### DIFF
--- a/CedarJavaFFI/src/interface.rs
+++ b/CedarJavaFFI/src/interface.rs
@@ -715,6 +715,10 @@ mod jvm_based_tests {
     static JVM: LazyLock<JavaVM> = LazyLock::new(|| create_jvm().unwrap());
 
     mod policy_tests {
+        use std::result;
+
+        use cedar_policy::Effect;
+
         use super::*;
 
         #[track_caller]
@@ -802,8 +806,202 @@ mod jvm_based_tests {
                 "myAnnotatedValue",
             );
         }
+        #[test]
+        fn test_parse_policy_internal_valid(){
+            let mut env = JVM.attach_current_thread().unwrap();
+            let input = r#"permit(principal,action,resource);"#;
+           let policy_jstr = env.new_string(input).unwrap();
+           let result = parse_policy_internal(&mut env, policy_jstr);
+           assert!(result.is_ok(),"Expected parse_policy_internal succeeded");
+
+           let jvalue = result.unwrap();
+           let parsed_jstring = JString::cast(&mut env, jvalue.l().unwrap()).unwrap();
+           let parsed_string = String::from(env.get_string(&parsed_jstring).unwrap());
+
+           assert!(parsed_string.contains("permit"));
+           assert!(parsed_string.contains("principal"))
+        }
+        
+        #[test]
+        fn parse_policy_internal_invalid()
+        {
+            let mut env = JVM.attach_current_thread().unwrap();
+            let input_invalid = r#"permit(Principal Action Resource);"#;
+            let jstr = env.new_string(input_invalid).unwrap();
+            let result = parse_policy_internal(&mut env, jstr);
+            if let Err(e) = &result {
+                println!("expected parsing failed")
+            }
+            assert!(result.is_err());
+        }
+
+       
+        #[test]
+        fn test_parse_policies_internal_invalid(){
+            let mut env = JVM.attach_current_thread().unwrap();
+
+            let invalid_input= "not a valid input";
+            let policy_jstr = env.new_string(invalid_input).unwrap();
+
+            let result = parse_policies_internal(&mut env, policy_jstr);
+            assert!(result.is_err(),"Expected to fail or invalid input");
+        }
+         #[test]
+         fn parse_policy_template_valid_test()
+         {
+             let mut env = JVM.attach_current_thread().unwrap();
+             let policy_template =  r#"permit(principal==?principal,action == Action::"readfile",resource==?resource );"#;
+             let jstr = env.new_string(policy_template).unwrap();
+            let result =parse_policy_template_internal (&mut env, jstr);
+            assert!(result.is_ok());
+         }
+
+         #[test]
+         fn parse_policy_template_invalid_test(){
+            let mut env = JVM.attach_current_thread().unwrap();
+            let invalid_input =  r#"permit(Principa,Action,Resource );"#; 
+            let jstr = env.new_string(invalid_input).unwrap();
+            let result = parse_policy_template_internal(&mut env, jstr);
+
+            if let Err(_e) = &result{
+                println!("expected failure{}",_e);
+            }
+
+            assert!(result.is_err())
+         }
+
+         #[test]
+        fn from_json_test_valid() {
+        let mut env = JVM.attach_current_thread().unwrap();
+
+        let policy_json = r#"
+        {
+            "effect": "permit",
+            "principal": {
+                "op": "==",
+                "entity": { "type": "User", "id": "12UA45" }
+            },
+            "action": {
+                "op": "==",
+                "entity": { "type": "Action", "id": "view" }
+            },
+            "resource": {
+                "op": "in",
+                "entity": { "type": "Folder", "id": "abc" }
+            },
+            "conditions": [
+                {
+                    "kind": "when",
+                    "body": {
+                        "==": {
+                            "left": {
+                                ".": {
+                                    "left": {
+                                        "Var": "context"
+                                    },
+                                    "attr": "tls_version"
+                                }
+                            },
+                            "right": {
+                                "Value": "1.3"
+                            }
+                        }
+                    }
+                }
+            ]
+        }
+            "#;
+
+    let jstr = env.new_string(policy_json).unwrap();
+    let result = from_json_internal(&mut env, jstr);
+
+    assert!(result.is_ok(), "Parsing failed: {:?}", result);
+
+    let jval = result.unwrap();
+    let obj = jval.l().unwrap();
+    let str_val: String = env.get_string(&obj.into()).unwrap().into();
+
+    println!("Returned policy string: '{}'", str_val);
+
+    assert!(
+        str_val.to_lowercase().contains("permit"),
+        "Expected 'permit' in the policy string, got: '{}'",
+        str_val
+    );
+}
+ #[test]
+    fn from_json_invalid()
+    {
+        let mut env = JVM.attach_current_thread().unwrap();
+        let invalid_input = r#"
+        {
+            "Effect": "permit",
+            "Principal": {
+                "op": "==",
+                "Entity": { "type": "User", "id": "12UA45" }
+            },
+            "Action": {
+                "op": "==",
+                "entity": { "type": "Action", "id": "view" }
+            },
+            "Resource": {
+                "op": "in",
+                "entity": { "type": "Folder", "id": "abc" }
+            },
+            "Conditions": [
+                {
+                    "kind": "when",
+                    "body": {
+                        "==": {
+                            "left": {
+                                ".": {
+                                    "left": {
+                                        "Var": "context"
+                                    },
+                                    "attr": "tls_version"
+                                }
+                            },
+                            "right": {
+                                "Value": "1.3"
+                            }
+                        }
+                    }
+                }
+            ]
+        }
+        "#;
+
+        let jstr = env.new_string(invalid_input).unwrap();
+        let result = from_json_internal(&mut env, jstr);
+        assert!(result.is_err(),"Not valid jason ");
     }
 
+   
+
+    #[test]
+    fn to_json_internal_test()
+    {
+     let mut env = JVM.attach_current_thread().unwrap();
+     let input = r#"permit(principal, action, resource);"#;
+     let jstr = env.new_string(input).unwrap();
+     let result = to_json_internal(&mut env, jstr);
+
+     assert!(result.is_ok(),"Valid json");
+
+    }
+
+    #[test]
+    fn to_json_internal_invalid()
+    {
+        let mut env = JVM.attach_current_thread().unwrap();
+        let invalid_input = r#"Permit(Principal, Resource, Action);"#;
+        let jstr = env.new_string(invalid_input).unwrap();
+
+        let result = from_json_internal(&mut env, jstr);
+        assert!(result.is_err(),"invalid input");
+    }
+        
+    }
     mod map_tests {
         use super::*;
 
@@ -820,6 +1018,7 @@ mod jvm_based_tests {
                 "Object should be a HashMap instance."
             );
         }
+        
 
         #[test]
         fn map_put_tests() {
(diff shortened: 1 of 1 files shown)


CedarJavaFFI/src/interface.rs
```diff
@@ -802,8 +806,192 @@ mod jvm_based_tests {
                 "myAnnotatedValue",
             );
         }
-    }
+        #[test]
+        fn test_parse_policy_internal_valid() {
```
mark-creamer-amazon, line 810: Nit: I think we should stay consistent with the test names. Imo we can remove the `test_` prefix. Here and everywhere else, e.g. lines `838`, `1067`, ...

CedarJavaFFI/src/interface.rs
```diff
@@ -802,8 +806,192 @@ mod jvm_based_tests {
...
+        #[test]
+        fn parse_policy_internal_invalid() {
+            let mut env = JVM.attach_current_thread().unwrap();
+            let input_invalid = r#"permit(Principal Action Resource);"#;
+            let jstr = env.new_string(input_invalid).unwrap();
+            let result = parse_policy_internal(&mut env, jstr);
+            if let Err(e) = &result {
+                println!("expected parsing failed")
+            }
+            assert!(result.is_err());
```
mark-creamer-amazon, line 834: Because we expect to fail here, we don't need to print in the case we do fail, so we can get rid of the `if let Err(e)...`

CedarJavaFFI/src/interface.rs
```diff
@@ -802,8 +806,192 @@ mod jvm_based_tests {
...
+
+        #[test]
+        fn test_parse_policies_internal_invalid() {
+            let mut env = JVM.attach_current_thread().unwrap();
+
+            let invalid_input = "not a valid input";
+            let policy_jstr = env.new_string(invalid_input).unwrap();
+
+            let result = parse_policies_internal(&mut env, policy_jstr);
+            assert!(result.is_err(), "Expected to fail or invalid input");
```
mark-creamer-amazon, line 845: Nit, could change to `Expected to fail for invalid input`. Also I think the fallback error message like this for the assertion is good, we should try and apply it consistently imo across the tests. E.g. line 834, 867, etc...

CedarJavaFFI/src/interface.rs
```diff
@@ -802,8 +806,192 @@ mod jvm_based_tests {
...
+        #[test]
+        fn parse_policy_template_invalid_test() {
+            let mut env = JVM.attach_current_thread().unwrap();
+            let invalid_input = r#"permit(Principa,Action,Resource );"#;
+            let jstr = env.new_string(invalid_input).unwrap();
+            let result = parse_policy_template_internal(&mut env, jstr);
+
+            if let Err(_e) = &result {
+                println!("expected failure{}", _e);
+            }
```
mark-creamer-amazon, line 865: Like with the above, we can get rid of this `if` and we don't need to print the error that we expected.

Also as an aside, in Rust the `_` before the variable name indicates that the variable is unused, and lets the compiler know not to complain about it not being used in the program. But this is inapplicable here as it's written, because `_e` is used later in the `println`. Just as a heads up, but either way I think we can get rid of L863-865

CedarJavaFFI/src/interface.rs
```diff
@@ -802,8 +806,192 @@ mod jvm_based_tests {
...
+            let jstr = env.new_string(policy_json).unwrap();
+            let result = from_json_internal(&mut env, jstr);
+
+            assert!(result.is_ok(), "Parsing failed: {:?}", result);
+
+            let jval = result.unwrap();
+            let obj = jval.l().unwrap();
+            let str_val: String = env.get_string(&obj.into()).unwrap().into();
+
+            println!("Returned policy string: '{}'", str_val);
```
mark-creamer-amazon, line 921: Imo we should remove all debug `println!` statements from the tests.

CedarJavaFFI/src/interface.rs
```diff
@@ -871,4 +1059,227 @@ mod jvm_based_tests {
...
+                "AesourceTypes": ["File"]
+                }
+            }
+            }
+        }
+        }"#;
+
+            let jstr = env.new_string(invalid_input).unwrap();
+            let result = parse_json_schema_internal(&mut env, jstr);
+            assert!(result.is_err(), " Expected parsing has failed");
```
mark-creamer-amazon, line 1157: Nit: ` Expected parsing has failed` to `Expected parse_json_schema_internal to fail`

CedarJavaFFI/src/interface.rs
```diff
@@ -871,4 +1059,227 @@ mod jvm_based_tests {
...
+            assert_eq!(policy_set.templates().count(), 1);
+        }
+        #[test]
+        fn template_effect_jni_internal_permit_test() {
+            let mut env = JVM.attach_current_thread().unwrap();
+            let template_policy = r#"permit(principal==?principal,action == Action::"readfile",resource==?resource );"#;
+
+            let jstr = env.new_string(template_policy).unwrap();
+            let result = template_effect_jni_internal(&mut env, jstr);
+            assert!(result.is_ok());
```
mark-creamer-amazon, line 1257: I think we could do some additional conversions on `result` to take the `JValueOwned` and turn it into a `&str` we could assert is `permit` in this case. Similar to how we do in `test_parse_cedar_schema_internal_valid` from lines `1198-1200`.

CedarJavaFFI/src/interface.rs
```diff
@@ -871,4 +1059,227 @@ mod jvm_based_tests {
...
+            let result = PolicySet::from_str(input);
+            assert!(
+                result.is_ok(),
+                "Expected template policy parsing to succeed"
+            );
+
+            let policy_set = result.unwrap();
+            assert_eq!(policy_set.policies().count(), 0);
+            assert_eq!(policy_set.templates().count(), 1);
+        }
```
mark-creamer-amazon, line 1249: Similar to the comment above, we can get rid of these tests as they're only testing already-covered methods from Cedar Rust and not using any from CedarJavaFFI

CedarJavaFFI/src/interface.rs
```diff
@@ -1118,21 +1166,23 @@ mod jvm_based_tests {
...
-            assert_eq!(rust_output, "success");
-        }
-
-        #[test]
-        fn parse_json_schema_internal_invalid_test() {
-            let mut env = JVM.attach_current_thread().unwrap();
-            let invalid_input = r#" {
+        let jstr = env.new_string(input).unwrap();
+        let result = parse_json_schema_internal(&mut env, jstr);
+        assert!(result.is_ok(), "successfully parsed json_schema_internal");
```
mcreamer: This `successfully parsed json_schema_internal` message would actually be printed when `result.is_ok()` is `false`, so we need to change this to something like `Expected schema to parse successfully` instead.

CedarJavaFFI/src/interface.rs
```diff
@@ -274,68 +274,67 @@ pub fn parsePoliciesJni<'a>(mut env: JNIEnv<'a>, _: JClass, policies_jstr: JStri
...
-        for template in policy_set.templates() {
-            let policy_id = format!("{}", template.id());
-            let policy_text = format!("{}", template);
-            let java_policy_object = JPolicy::new(
-                env,
-                &env.new_string(&policy_text)?,
-                &env.new_string(&policy_id)?,
-            )?;
-            let _ = templates_java_hash_set.add(env, java_policy_object);
+                let policy_set_obj = create_java_policy_set(env, policies_set.as_ref(), template_set.as_ref())?; //unknown
```
mcreamer: Nit: I think we can remove the `//unknown` comment

CedarJavaFFI/src/interface.rs
```diff
@@ -274,68 +274,67 @@ pub fn parsePoliciesJni<'a>(mut env: JNIEnv<'a>, _: JClass, policies_jstr: JStri
...
-            env,
-            policies_java_hash_set.as_ref(),
-            templates_java_hash_set.as_ref(),
-        );
-
-        Ok(JValueGen::Object(java_policy_set))
     }
 }
-
+    fn build_java_policy_set<'a>(env: &mut JNIEnv<'a>,entries: Vec<(String, String)>,)-> Result<Set<'a, JPolicy<'a>>>  {
```
mcreamer: I think this is a good helper method, but I think `build_java_policy_set` is a confusing name when there's the existing `create_java_policy_set` method below.

I think we could rename `build_java_policy_set` to `build_jpolicy_jset` or something similar to disambiguate with our `PolicySet` struct or analagous Java class

CedarJavaFFI/src/interface.rs
```diff
@@ -571,16 +578,21 @@ pub fn parseEntityTypeName<'a>(mut env: JNIEnv<'a>, _: JClass, obj: JString<'a>)
     }
 }
 
+fn parse_entity_type_name_str(input: &str) -> Result<EntityTypeName>
+{
+    EntityTypeName::from_str(input).map_err(|e| Box::new(e) as _)
```
mcreamer, line 589: Do we need the `as _`?

CedarJavaFFI/src/interface.rs
```diff
@@ -549,18 +548,26 @@ pub fn getEntityIdentifierRepr<'a>(mut env: JNIEnv<'a>, _: JClass, obj: JObject<
...
+        return raise_npe(env);
     }
+    //cast wrapper
+    let eid = JEntityId::cast(env, obj)?;
+
+    //Error handling
+    let repr = get_entity_identifier_str(&eid);
+    let jstring = env.new_string(&repr)?;
+    Ok(JValueGen::Object(jstring.into()))
+
```
mcreamer: I'd be ok with removing the `else` clause here, as long as it's applied consistently with the style elsewhere. For instance, I'd expect the change to be done in `parse_entity_type_name_internal` as well, but we left the `else` alone there.

I think imo we can just leave the `else` clause alone here.

CedarJavaFFI/src/interface.rs
```diff
@@ -691,19 +729,19 @@ fn policies_str_to_pretty_internal<'a>(
...
+        {
             Config::default()
         };
         let policies_str = String::from(env.get_string(&policies_jstr)?);
-        match policies_str_to_pretty(&policies_str, &config) {
-            Ok(formatted_policies) => Ok(env.new_string(formatted_policies)?.into()),
-            Err(e) => Err(e.into()),
-        }
+        let formatted_string = format_policies_str_with_config(&policies_str, &config)?;
+        Ok(env.new_string(formatted_string)?.into())
```
mcreamer, line 753: I think it's worth putting this in a `result` variable or similar

CedarJavaFFI/src/interface.rs
```diff
@@ -1058,48 +1092,62 @@ mod jvm_based_tests {
...
-                }
-            }
-            }
-        }
-        }"#;
-
-            let result = Schema::from_json_str(json);
-            assert!(result.is_ok(), "Expected schema parsing to succeed");
-        }
+        fn test_parse_cedar_schema_internal_invalid() {
```
mcreamer, line 1118: Nit remove `test_` prefix here

CedarJavaFFI/src/interface.rs
```diff
@@ -1152,103 +1202,90 @@ mod jvm_based_tests {
...
-            let jstr = env.new_string(invalid_input).unwrap();
-            let result = parse_json_schema_internal(&mut env, jstr);
-            assert!(result.is_err(), " Expected parsing has failed");
+        let jstr = env.new_string(invalid_input).unwrap();
+        let result = parse_json_schema_internal(&mut env, jstr);
+        assert!(result.is_err(), "Expected json_schema_internal parsing to fail: {:?}", result);     
+    }
+   
+         #[test]
+        fn test_parse_policies_set_text_map()
```
mcreamer: Nit remove `test_` prefix here

CedarJavaFFI/src/interface.rs
```diff
@@ -807,35 +845,40 @@ mod jvm_based_tests {
...
 
             let jvalue = result.unwrap();
             let parsed_jstring = JString::cast(&mut env, jvalue.l().unwrap()).unwrap();
-            let parsed_string = String::from(env.get_string(&parsed_jstring).unwrap());
-
-            assert!(parsed_string.contains("permit"));
-            assert!(parsed_string.contains("principal"))
+            let actual_parsed_string = String::from(env.get_string(&parsed_jstring).unwrap());
+            let expected_policy_object = cedar_policy::Policy::from_str(input).unwrap();
+            let expected_canonical_string = format!("{}", expected_policy_object);
```
mcreamer, line 875: We should be able to do `expected_policy_object.to_string()` instead, right?

CedarJavaFFI/src/interface.rs
```diff
@@ -1281,5 +1314,229 @@ mod jvm_based_tests {
...
+        let actual_repr = get_entity_identifier_string_representation(&entity_id);
+        assert_eq!(actual_repr, expected_repr);
+    }
+
+    #[test]
+    fn  get_entity_identifier_string_representation_long_string() {
+        let long_id = "a".repeat(1000); 
+        let entity_id = EntityId::new(long_id.clone());
+        
+        let expected_repr = format!("{}", long_id);
```
mcreamer: Do we need `format!("{}", long_id)` here? Can we do `long_id.to_string()`?

CedarJavaFFI/src/interface.rs
```diff
@@ -1281,5 +1314,229 @@ mod jvm_based_tests {
...
+        let input ="";
+        let result = parse_entity_type_name_str(input);
+        assert!(result.is_err(),"Expected parsing failed due to an empty string");
+    }
+    #[test]
+    fn parse_entity_type_name_separatorsstr()
+    {
+        let input = "::";
+        let result = parse_entity_type_name_str(input);
+        assert!(result.is_err(),"Expected parsing failed for only seperators");
```
mcreamer: Nit `seperators` to `separators`

CedarJavaFFI/src/interface.rs
swabyra: I had it removed but i think I didn't merge the new changes with the stash. I'll put the new refactor in the next PR.

CedarJavaFFI/src/interface.rs
```diff
@@ -988,9 +1022,9 @@ mod jvm_based_tests {
             let invalid_input = r#"Permit(Principal, Resource, Action);"#;
             let jstr = env.new_string(invalid_input).unwrap();
 
-            let result = from_json_internal(&mut env, jstr);
-            assert!(result.is_err(), "invalid input");
-        }
+        let result = from_json_internal(&mut env, jstr);
+        assert!(result.is_err(), "Expected json_internal parsing to fail: {:?}", result);    }
```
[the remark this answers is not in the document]
swabyra: this one also got mixed up in the selecting changes I believe will just put this one on the new PR as well.

CedarJavaFFI/src/interface.rs
swabyra: I can look into that.

CedarJavaFFI/src/interface.rs
```diff
@@ -856,30 +864,39 @@ mod jvm_based_tests {
...
+    let parsed_jstring = JString::cast(&mut env, jvalue.l().unwrap()).unwrap();
+    let actual_parsed_string = String::from(env.get_string(&parsed_jstring).unwrap());
+    let expected_policy_object = cedar_policy::Policy::from_str(input).unwrap();
+    let expected_canonical_string = expected_policy_object.to_string();
+
+    assert_eq!(
+        actual_parsed_string, expected_canonical_string,
+        "Parsed policy string should match the expected canonical format."
+    );
+}
```
mcreamer, line 896: Identation, I think we need to run `cargo +nightly fmt`?

CedarJavaFFI/src/interface.rs
```diff
@@ -1371,57 +1459,72 @@ mod jvm_based_tests {
...
         }
-    }
-
-    #[cfg(test)]
-    mod helper_tests {
-        use std::result;
+        #[test]
+        fn get_entity_identifier_repr_internal_raw_string_valid() {
+            let mut env = JVM.attach_current_thread().unwrap();
+            let input = env.new_string("alice").unwrap(); // Java string
```
mcreamer, line 1465: Nit: can get rid of this `// Java string` comment

CedarJavaFFI/src/interface.rs
```diff
@@ -1371,57 +1459,72 @@ mod jvm_based_tests {
...
+                }
+            }
+        }
+    }"#;
 
-            let expected_repr = format!("{}", long_id);
-            let actual_repr = get_entity_identifier_string_representation(&entity_id);
-            assert_eq!(actual_repr, expected_repr);
+            let result = validate_entities(input);
+            assert!(result.is_ok());
```
mcreamer, line 1519: Hmm, since `User::"bob"`'s `age` is `not a number` when it should be a `Long`, I think this should be failing, right?

CedarJavaFFI/src/interface.rs
swabyra: yeah I'll run it again on the next PR .

CedarJavaFFI/src/interface.rs
```diff
@@ -615,12 +611,11 @@ fn get_entity_type_name_repr_internal<'a>(
     obj: JObject<'a>,
 ) -> Result<JValueOwned<'a>> {
     if obj.is_null() {
-        raise_npe(env)
-    } else {
-        let etype = JEntityTypeName::cast(env, obj)?;
-        let repr = etype.get_string_repr();
-        Ok(env.new_string(repr)?.into())
+        let _ = raise_npe(env);
```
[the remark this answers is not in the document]
swabyra: fixed

CedarJavaFFI/src/interface.rs
```diff
@@ -1495,77 +1621,176 @@ mod jvm_based_tests {
...
+
+        #[test]
+        fn get_entity_type_name_repr_null() {
+            let mut env = JVM.attach_current_thread().unwrap();
+            let null_obj = JObject::null();
+            let result = get_entity_type_name_repr_internal(&mut env, null_obj);
+            assert!(
+                result.is_err(),
+                "Expected get_entity_type_name to fail for null input"
+            );
```
mcreamer, line 1776: Different `assert!` pattern here for the null tests, should revisit here when the dust settles on how we're proceeding with these cases

CedarJavaFFI/src/interface.rs
```diff
@@ -586,20 +583,19 @@ pub fn parseEntityTypeName<'a>(mut env: JNIEnv<'a>, _: JClass, obj: JString<'a>)
 }
 
 fn parse_entity_type_name_str(input: &str) -> Result<EntityTypeName> {
-    EntityTypeName::from_str(input).map_err(|e| Box::new(e) as _)
+    Ok(EntityTypeName::from_str(input).map_err(|e| Box::new(e))?)
```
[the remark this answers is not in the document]
swabyra: I tried something a bit different, will be in the next pr.

CedarJavaFFI/src/interface.rs
```diff
@@ -785,6 +786,13 @@ mod jvm_based_tests {
             policy_effect_test_util(&mut env, "permit(principal,action,resource);", "permit");
             policy_effect_test_util(&mut env, "forbid(principal,action,resource);", "forbid");
         }
+        #[test]
+        fn policy_effect_jni_internal_null() {
+            let mut env = JVM.attach_current_thread().unwrap();
+            let null_obj = JObject::null();
+            let result = policy_effect_jni_internal(&mut env, null_obj.into());
+            assert!(result.is_ok(), "Expected error on null input");
```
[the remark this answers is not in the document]
swabyra: I ended up trying the env.exception

CedarJavaFFI/src/interface.rs
```diff
@@ -597,12 +611,11 @@ fn get_entity_type_name_repr_internal<'a>(
...
 ) -> Result<JValueOwned<'a>> {
     if obj.is_null() {
-        raise_npe(env)
-    } else {
-        let etype = JEntityTypeName::cast(env, obj)?;
-        let repr = etype.get_string_repr();
-        Ok(env.new_string(repr)?.into())
+        let _ = raise_npe(env);
     }
+    let jstr = JString::from(obj);
```
muditchaudhary, line 616: I'm a bit unsure about this refactor. The corresponding CedarJava method is `getEntityTypeNameRepr(EntityTypeName type)`. So, it sends a `EntityTypeName` object as an input. We are trying to get a `JString` from `EntityTypeName` which makes me think that it will fail at runtime. Can we test this method by calling it from CedarJava to verify?

CedarJavaFFI/src/interface.rs
```diff
@@ -613,19 +626,43 @@ pub fn parseEntityUID<'a>(mut env: JNIEnv<'a>, _: JClass, obj: JString<'a>) -> j
...
     obj: JString<'a>,
 ) -> Result<JValueOwned<'a>> {
     if obj.is_null() {
-        raise_npe(env)
-    } else {
-        let jstring = env.get_string(&obj)?;
-        let src = String::from(jstring);
-        let obj = JEntityUID::parse(env, &src)?;
-        Ok(obj.into())
+        return raise_npe(env);
```
muditchaudhary, line 644: We don't need `;`

CedarJavaFFI/src/interface.rs
```diff
@@ -613,19 +626,43 @@ pub fn parseEntityUID<'a>(mut env: JNIEnv<'a>, _: JClass, obj: JString<'a>) -> j
...
+    euid_str: &str,
+) -> Result<HashMap<String, String>> {
+    let cedar_euid = EntityUid::from_str(euid_str)?;
+    let mut result = HashMap::new();
+    result.insert("id".to_string(), format!("{:?}", cedar_euid.type_id()));
+    result.insert("type".to_string(), cedar_euid.type_name().to_string());
+    Ok(result)
+}
 
 fn parse_entity_uid_internal<'a>(
```
muditchaudhary, line 639: We are changing FFI's contract with CedarJava here. Currently, the corresponding CedarJava method `private static native Optional<EntityUID> parseEntityUID(String src)` is expecting a return of type `Optional<EntityUID`. Here, we are updating it to `Map<String,String>`. But because we haven't updated the method in CedarJava, calling this method will fail.

CedarJavaFFI/src/interface.rs
```diff
@@ -644,18 +681,26 @@ pub fn getEUIDRepr<'a>(
...
-        let id = JEntityId::cast(env, id)?.get_rust_repr();
-        let euid = EntityUid::from_type_name_and_id(etype, id);
-        let jstring = env.new_string(euid.to_string())?;
-        Ok(jstring.into())
+    if type_name_obj.is_null() || id_obj.is_null() {
+        return raise_npe(env);
     }
+
+    // Convert JObject -> JString
+    let type_jstr = JString::from(type_name_obj);
```
muditchaudhary, line 692: Corresponding CedarJava method `private static native String getEUIDRepr(EntityTypeName type, EntityIdentifier id)`. I believe this would fail as the inputs are of types `EntityTypeName` and `EntityIdentifier`. Can we test this through CedarJava to verify?

CedarJavaFFI/src/interface.rs
```diff
@@ -683,6 +728,14 @@ pub fn policiesStrToPrettyWithConfig<'a>(
     }
 }
 
+fn format_policies_str_with_config(
```
muditchaudhary, line 731: Do we want this as a separate method as it is a single line?

CedarJavaFFI/src/interface.rs
```diff
@@ -802,8 +863,243 @@ mod jvm_based_tests {
...
+            let result = parse_policy_internal(&mut env, null_str);
+            assert!(result.is_ok(), "Expected error on null input");
+        }
+        #[test]
+        fn parse_policy_template_valid_test() {
+            let mut env = JVM.attach_current_thread().unwrap();
+            let policy_template = r#"permit(principal==?principal,action == Action::"readfile",resource==?resource );"#;
+            let jstr = env.new_string(policy_template).unwrap();
+            let result = parse_policy_template_internal(&mut env, jstr);
+            assert!(result.is_ok());
```
muditchaudhary, line 941: Can we also verify the value of parsed policy template like we did in `parse_policy_internal_valid`

CedarJavaFFI/src/interface.rs
```diff
@@ -802,8 +863,243 @@ mod jvm_based_tests {
...
+            assert!(
+                result.is_ok(),
+                "Expected from_json parsing to succeed, got: {:?}",
+                result
+            );
 
+            let jval = result.unwrap();
+            let obj = jval.l().unwrap();
+            let str_val: String = env.get_string(&obj.into()).unwrap().into();
+            assert!(
```
muditchaudhary, line 1008: We should assert with the equivalent cedar policy instead of just the effect

CedarJavaFFI/src/interface.rs
```diff
@@ -802,8 +863,243 @@ mod jvm_based_tests {
...
+        }
+
+        #[test]
+        fn to_json_internal_test() {
+            let mut env = JVM.attach_current_thread().unwrap();
+            let input = r#"permit(principal, action, resource);"#;
+            let jstr = env.new_string(input).unwrap();
+            let result = to_json_internal(&mut env, jstr);
+
+            assert!(result.is_ok(), "Valid json");
```
muditchaudhary, line 1079: We should assert with a value instead of just `is_ok()`

CedarJavaFFI/src/interface.rs
```diff
@@ -870,5 +1166,631 @@ mod jvm_based_tests {
                 "Retrieved value should be equal to the inserted value."
             )
         }
+        #[test]
```
muditchaudhary, line 1169: These tests don't belong under `mod map_tests`

CedarJavaFFI/src/interface.rs
```diff
@@ -870,5 +1166,631 @@ mod jvm_based_tests {
...
+
+            assert_eq!(output_str, r#""alice""#);
+        }
+        #[test]
+        fn get_entity_identifier_repr_internal_with_escape() {
+            let mut env = JVM.attach_current_thread().unwrap();
+            let input = env.new_string(r#"bob\smith"#).unwrap(); // backslash
+
+            let result_jval =
+                get_entity_identifier_repr_internal(&mut env, JObject::from(input)).unwrap();
```
muditchaudhary, line 1480: The corresponding CedarJava method is `String getEntityIdentifierRepr(EntityIdentifier id)` but we are passing a String here as input to `get_entity_identifier_repr_internal`

muditchaudhary (changes requested):
Can we break it into 2 separate PRs? 1) Adding tests for methods that don't need a refactor 2) Refactored methods and their tests

CedarJavaFFI/src/interface.rs
swabyra: fixed

PR #349, opened 2026-04-16
Optimize JNI performance and fix Decimal validation

geekphilosophy:
Improve JNI call performance ~3x with reduced overhead. Add JMH benchmark suite for tracking JNI performance. Fix Decimal validator to accept leading zeros in corpus integration tests.

*Issue #, if available:*

*Description of changes:*
Cedar-Java JMH Benchmark Results — After Performance Fixes
============================================================
Date:    2026-04-15
Base:    1d141bc (main)
JDK:     21.0.10 (Corretto)
JMH:     1.37
OS:      macOS (Darwin 25.3.0, aarch64)
Config:  @Warmup(iterations=3, time=1) @Measurement(iterations=5, time=1) @Fork(1)

Changes applied:
  1. Remove thread::spawn per JNI call, use std::panic::catch_unwind instead (interface.rs)
  2. Remove redundant j_input_str.push(' ') (interface.rs)
  3. Eliminate double JSON deserialization (readTree+readValue → single readValue) (BasicAuthorizationEngine.java)
  4. Cache JNI version check in static initializer instead of per-call (BasicAuthorizationEngine.java)
  5. Single-pass escape detection in ValueDeserializer using node.size() + node.has() (ValueDeserializer.java)

Results (thorough run: 2 forks, 5 warmup, 20 measurement iterations):
Benchmark                                   Mode  Cnt    Score    Error  Units
AuthorizationBenchmark.isAuthorized_medium  avgt   40  122.606 ±  2.303  us/op
AuthorizationBenchmark.isAuthorized_small   avgt   40   25.416 ±  0.837  us/op
AuthorizationBenchmark.validate_small       avgt   40  159.865 ± 29.990  us/op

Comparison to baseline (1d141bc):
                          Baseline        After         Improvement
  isAuthorized_small      85.4 ± 3.4    25.4 ± 0.8     ~3.4x faster
  isAuthorized_medium    329.1 ± 258.4  122.6 ± 2.3    ~2.7x faster
  validate_small         353.3 ± 146.0  159.9 ± 30.0   ~2.2x faster

### DIFF
--- a/CedarJava/src/jmh/java/com/cedarpolicy/AuthorizationBenchmark.java
+++ b/CedarJava/src/jmh/java/com/cedarpolicy/AuthorizationBenchmark.java
@@ -173,17 +173,17 @@ public class AuthorizationBenchmark {
     }
 
     @Benchmark
-    public AuthorizationResponse isAuthorized_small() throws AuthException {
+    public AuthorizationResponse isAuthorizedSmall() throws AuthException {
         return engine.isAuthorized(smallRequest, smallPolicySet, smallEntities);
     }
 
     @Benchmark
-    public AuthorizationResponse isAuthorized_medium() throws AuthException {
+    public AuthorizationResponse isAuthorizedMedium() throws AuthException {
         return engine.isAuthorized(mediumRequest, mediumPolicySet, mediumEntities);
     }
 
     @Benchmark
-    public ValidationResponse validate_small() throws AuthException {
+    public ValidationResponse validateSmall() throws AuthException {
         return engine.validate(validationRequest);
     }
 }
--- a/CedarJava/src/jmh/java/com/cedarpolicy/AuthorizationBenchmark.java
+++ b/CedarJava/src/jmh/java/com/cedarpolicy/AuthorizationBenchmark.java
@@ -1,0 +1,189 @@
+/*
+ * Copyright Cedar Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.cedarpolicy;
+
+import com.cedarpolicy.model.AuthorizationRequest;
+import com.cedarpolicy.model.AuthorizationResponse;
+import com.cedarpolicy.model.ValidationRequest;
+import com.cedarpolicy.model.ValidationResponse;
+import com.cedarpolicy.model.entity.Entity;
+import com.cedarpolicy.model.exception.AuthException;
+import com.cedarpolicy.model.policy.Policy;
+import com.cedarpolicy.model.policy.PolicySet;
+import com.cedarpolicy.model.schema.Schema;
+import com.cedarpolicy.model.schema.Schema.JsonOrCedar;
+import com.cedarpolicy.value.EntityTypeName;
+import com.cedarpolicy.value.EntityUID;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * JMH benchmarks for Cedar authorization engine JNI performance.
+ *
+ * <p>Run via: ./gradlew jmh
+ */
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@Warmup(iterations = 3, time = 1)
+@Measurement(iterations = 5, time = 1)
+@Fork(1)
+@State(Scope.Benchmark)
+public class AuthorizationBenchmark {
+
+    private BasicAuthorizationEngine engine;
+
+    // Small scenario: minimal request
+    private AuthorizationRequest smallRequest;
+    private PolicySet smallPolicySet;
+    private Set<Entity> smallEntities;
+
+    // Medium scenario: photoflash schema with entities and multiple policies
+    private AuthorizationRequest mediumRequest;
+    private PolicySet mediumPolicySet;
+    private Set<Entity> mediumEntities;
+
+    // Validation scenario
+    private ValidationRequest validationRequest;
+
+    @Setup(Level.Trial)
+    public void setUp() throws Exception {
+        engine = new BasicAuthorizationEngine();
+
+        setUpSmallScenario();
+        setUpMediumScenario();
+        setUpValidationScenario();
+    }
+
+    private void setUpSmallScenario() {
+        EntityTypeName userType = EntityTypeName.parse("User").get();
+        EntityTypeName actionType = EntityTypeName.parse("Action").get();
+        EntityTypeName resourceType = EntityTypeName.parse("Resource").get();
+
+        EntityUID principal = userType.of("alice");
+        EntityUID action = actionType.of("view");
+        EntityUID resource = resourceType.of("doc1");
+
+        smallRequest = new AuthorizationRequest(principal, action, resource, new HashMap<>());
+
+        Set<Policy> policies = new HashSet<>();
+        policies.add(new Policy("permit(principal, action, resource);", "policy0"));
+        smallPolicySet = new PolicySet(policies);
+
+        smallEntities = new HashSet<>();
+        smallEntities.add(new Entity(principal, new HashMap<>(), new HashSet<>()));
+        smallEntities.add(new Entity(action, new HashMap<>(), new HashSet<>()));
+        smallEntities.add(new Entity(resource, new HashMap<>(), new HashSet<>()));
+    }
+
+    private void setUpMediumScenario() throws Exception {
+        EntityTypeName userType = EntityTypeName.parse("User").get();
+        EntityTypeName actionType = EntityTypeName.parse("Action").get();
+        EntityTypeName photoType = EntityTypeName.parse("Photo").get();
+        EntityTypeName albumType = EntityTypeName.parse("Album").get();
+        EntityTypeName accountType = EntityTypeName.parse("Account").get();
+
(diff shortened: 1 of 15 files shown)


CedarJava/src/jmh/java/com/cedarpolicy/AuthorizationBenchmark.java
```diff
@@ -0,0 +1,189 @@
...
+        policies.add(new Policy("permit(principal, action, resource);", "policy0"));
+        smallPolicySet = new PolicySet(policies);
+
+        smallEntities = new HashSet<>();
+        smallEntities.add(new Entity(principal, new HashMap<>(), new HashSet<>()));
+        smallEntities.add(new Entity(action, new HashMap<>(), new HashSet<>()));
+        smallEntities.add(new Entity(resource, new HashMap<>(), new HashSet<>()));
+    }
+
+    private void setUpMediumScenario() throws Exception {
```
mark-creamer-amazon, line 107: Nit: I think if we anticipate adding more scenarios, we might want to rename this (perhaps `basicPhotoFlashScenario`), but I think this is good as is

CedarJava/src/main/java/com/cedarpolicy/serializer/ValueDeserializer.java
```diff
@@ -85,68 +85,47 @@ public Value deserialize(JsonParser parser, DeserializationContext context) thro
...
-                                    } else {
-                                        String msg = "Invalid Entity Type" + val.get("type").textValue();
-                                        throw new InvalidValueDeserializationException(parser, msg, node.asToken(), Map.class);
-                                    }
-                                }
+                // Single-pass: check for escape sequences via direct lookup, avoiding full iteration
+                if (node.size() == 1) {
+                    if (node.has(ENTITY_ESCAPE_SEQ)) {
+                        JsonNode val = node.get(ENTITY_ESCAPE_SEQ);
+                        if (val.isObject() && val.has("id") && val.has("type") && val.size() == 2) {
```
mark-creamer-amazon: Nit: Looks like we can also shed the `val.isObject() &&` because [`val.has(...)`](https://www.javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/latest/com/fasterxml/jackson/databind/JsonNode.html#has-java.lang.String-) checks this already

> Method that allows checking whether this node is JSON Object node and contains value for specified property.

CedarJava/src/main/java/com/cedarpolicy/serializer/ValueDeserializer.java
```diff
@@ -85,68 +85,47 @@ public Value deserialize(JsonParser parser, DeserializationContext context) thro
...
-                                    EntityIdentifier id = new EntityIdentifier(val.get("id").textValue());
-                                    Optional<EntityTypeName> type = EntityTypeName.parse(val.get("type").textValue());
-                                    if (type.isPresent()) {
-                                        return new EntityUID(type.get(), id);
-                                    } else {
-                                        String msg = "Invalid Entity Type" + val.get("type").textValue();
-                                        throw new InvalidValueDeserializationException(parser, msg, node.asToken(), Map.class);
-                                    }
-                                }
+                // Single-pass: check for escape sequences via direct lookup, avoiding full iteration
```
mark-creamer-amazon, line 82: Nice refactor here! Deserializer changes look equivalent to me

CedarJava/src/main/java/com/cedarpolicy/serializer/ValueDeserializer.java
muditchaudhary: Nit: We can get rid of `EscapeType` as it is not longer being used.

CedarJava/src/main/java/com/cedarpolicy/serializer/ValueDeserializer.java
muditchaudhary: +1

CedarJava/src/main/java/com/cedarpolicy/serializer/ValueDeserializer.java
```diff
@@ -85,68 +85,47 @@ public Value deserialize(JsonParser parser, DeserializationContext context) thro
...
+                if (node.size() == 1) {
+                    if (node.has(ENTITY_ESCAPE_SEQ)) {
+                        JsonNode val = node.get(ENTITY_ESCAPE_SEQ);
+                        if (val.isObject() && val.has("id") && val.has("type") && val.size() == 2) {
+                            EntityIdentifier id = new EntityIdentifier(val.get("id").textValue());
+                            Optional<EntityTypeName> type = EntityTypeName.parse(val.get("type").textValue());
```
muditchaudhary, line 88: Not required for this PR, but something I remembered while reviewing it: we can further optimize the deserializer by caching results from pure functions that cross the FFI boundary like `EntityTypeName.parse(_)` to avoid redundant and expensive FFI roundtrip costs for the same inputs.  Can probably extend this idea beyond deserializers.

CedarJava/src/main/java/com/cedarpolicy/BasicAuthorizationEngine.java
```diff
@@ -42,14 +42,20 @@
...
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JsonNode;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /** An authorization engine that is compiled in process. Communicated with via JNI. */
 public final class BasicAuthorizationEngine implements AuthorizationEngine {
     static {
         LibraryLoader.loadLibrary();
+        String jniVersion = getCedarJNIVersion();
```
muditchaudhary, line 52: Nice, this makes perfect sense. The per-call version check in `call` would only be necessary if the FFI library wasn't statically loaded. Since it is, checking once at initialization is sufficient.

CedarJava/src/main/java/com/cedarpolicy/value/Decimal.java
```diff
@@ -36,9 +36,6 @@ public static boolean validDecimal(String d) {
                 return false;
             }
             d = d.trim();
-            if (d.length() > 21) {
```
muditchaudhary, line 39: This check oversimplified the validation. 

> A `decimal` value can range from `-922337203685477.5808` to `922337203685477.5807.`
([source](https://docs.cedarpolicy.com/policies/syntax-datatypes.html#datatype-decimal))

But Cedar trims leading zeroes. So a value like `-000000922337203685477.5808` is perfectly valid (Checked on [Cedar policy playground](https://www.cedarpolicy.com/en/playground))  yet would fail `d.length() > 21`.

This went uncaught for ~2 years and only surfaced with the new corpus tests. Rather than just removing the check, we should fix it to correctly validate the range after stripping leading zeroes.

CedarJava/src/jmh/java/com/cedarpolicy/AuthorizationBenchmark.java
```diff
@@ -0,0 +1,189 @@
...
+        EntityTypeName accountType = EntityTypeName.parse("Account").get();
+
+        EntityUID principal = userType.of("alice");
+        EntityUID action = actionType.of("View_Photo");
+        EntityUID resource = photoType.of("pic01");
+
+        mediumRequest = new AuthorizationRequest(principal, action, resource, new HashMap<>());
+
+        Set<Policy> policies = new HashSet<>();
+        policies.add(new Policy(
```
muditchaudhary: Nit: We can put the policies in a file and use `PolicySet.parsePolicies()` [Example](https://github.com/cedar-policy/cedar-java/blob/main/CedarJava/src/test/java/com/cedarpolicy/PolicySetTests.java#L40)

Same for Entities using `Entities.parse` [Example](https://github.com/cedar-policy/cedar-java/blob/main/CedarJava/src/test/java/com/cedarpolicy/EntitiesTests.java#L124)

CedarJavaFFI/src/interface.rs
```diff
@@ -82,17 +78,14 @@ pub fn callCedarJNI(
...
+    let j_input_str: String = match env.get_string(&j_input) {
         Ok(s) => s.into(),
         Err(_) => return build_err_obj(&env, "parsing"),
     };
-    j_input_str.push(' ');
-
-    let handle = thread::spawn(move || call_cedar_in_thread(j_call_str, j_input_str));
 
-    let result = match handle.join() {
+    let result = match panic::catch_unwind(|| call_cedar(&j_call_str, &j_input_str)) {
```
muditchaudhary, line 86: This is nice.

Nit: `call_cedar_in_thread` is probably dead code now.

muditchaudhary (changes requested):
Mostly nits except for decimal validation.

CedarJava/src/main/java/com/cedarpolicy/value/Decimal.java
geekphilosophy: Done, added specific tests for this.

CedarJava/src/test/java/com/cedarpolicy/DecimalTests.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+package com.cedarpolicy;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+import com.cedarpolicy.value.Decimal;
+
+public class DecimalTests {
```
muditchaudhary, line 24: We should also add tests for weird edge cases like `.`, `1.`, `-.0`, `1.2.3`, `-.`. Going through the decimal validator, I feel that some of these might not throw an exception and will require tightening up the regex.

CedarJava/src/main/java/com/cedarpolicy/value/Decimal.java
```diff
@@ -30,21 +29,48 @@ public class Decimal extends Value {
...
+            String withoutSign = negative ? d.substring(1) : d;
+
+            int dotIndex = withoutSign.indexOf('.');
+            String intPart = withoutSign.substring(0, dotIndex);
+            String fracPart = withoutSign.substring(dotIndex + 1);
+
+            // Strip leading zeros from integer part
+            intPart = intPart.replaceFirst("^0+", "");
+            if (intPart.isEmpty()) {
+                intPart = "0";
```
muditchaudhary: This will lead to accepting cases like `.1` which is invalid. Can tighten the upstream regex and not have this conditional instead.

CedarJava/src/main/java/com/cedarpolicy/value/Decimal.java
```diff
@@ -30,21 +29,48 @@ public class Decimal extends Value {
...
+            // Pad fractional part to exactly 4 digits
+            fracPart = fracPart + "0000".substring(fracPart.length());
+
+            // Combine integer and fractional parts as the i64 representation (value * 10000)
+            String combined = intPart.equals("0") ? fracPart.replaceFirst("^0+", "") : intPart + fracPart;
+            if (combined.isEmpty()) {
+                combined = "0";
+            }
+
+            String limit = negative ? MAX_NEGATIVE : MAX_POSITIVE;
```
muditchaudhary: Possibly range validation can be simplified using `BigDecimal`. 
```
BigDecimal val = new BigDecimal(d);
return val.compareTo(RANGE_MIN) >= 0 && val.compareTo(RANGE_MAX) <= 0;
```

Rest of the PR looks all good to me.

CedarJava/src/test/java/com/cedarpolicy/DecimalTests.java
geekphilosophy: Done, with a few more.

CedarJava/src/main/java/com/cedarpolicy/value/Decimal.java
geekphilosophy: Done, great simplification.

CedarJava/src/main/java/com/cedarpolicy/value/Decimal.java
geekphilosophy: Fixed.

muditchaudhary (approved):

mark-creamer-amazon (approved):

PR #350, opened 2026-04-25
Add policy and schema caching via cache()/cacheKey() API

geekphilosophy:
Adds Rust-side caching of pre-parsed policies and schemas, integrated directly into PolicySet and Schema. When cached, BasicAuthorizationEngine transparently uses a fast stateful authorization path that skips re-parsing — no separate engine class or new authorization flow required.

Usage:

  policySet.cache();
  schema.cache();

  var engine = new BasicAuthorizationEngine();
  for (AuthorizationRequest req : requests) {
      engine.isAuthorized(req, policySet, entities);  // fast cached path
  }
  // Cached data is freed automatically when the object is GC'd.

API on PolicySet and Schema:
- cache(): pre-parse and cache on the Rust side. One-way, idempotent. To use different policies/schema, create a new instance.
- cacheKey(): returns Optional<String> cache ID, empty if not cached.

Both policy set and schema must be cached for the fast path to be used. If only one is cached, authorization falls back to the uncached path to avoid silently skipping schema validation.

Implementation:

Rust (CedarJavaFFI):
- Shared DashMap cache for parsed PolicySets and Schemas, accessible from any thread. Direct JNI entry points for cache removal (no JSON dispatch overhead).
- Stateful authorization operation that looks up cached policies/schema by ID and authorizes against them.

Java:
- PolicySet and Schema gain cache()/cacheKey() methods.
- BasicAuthorizationEngine.isAuthorized() checks for cached inputs and dispatches to the stateful path transparently.
- GC-based cleanup via java.lang.ref.Cleaner frees Rust cache entries.
- SharedCedarInternals centralizes the Cleaner instance and JNI access. NativeHelpers remains package-private.

Benchmarks:
- Full cross-product matrix: 4 policy sizes x 3 entity sizes, cached and uncached. Run via: ./gradlew jmh
- Caching avoids re-parsing policies and schemas on every authorization call. The benefit scales with policy/schema complexity — from ~2x for trivial policies to over 20x for large policy sets with schemas


CedarJava/build.gradle
```diff
@@ -301,8 +301,8 @@ tasks.register('jmh', JavaExec) {
 }
 
 compileJava {
-    sourceCompatibility = "1.8"
-    targetCompatibility = "1.8"
+    sourceCompatibility = "17"
+    targetCompatibility = "17"
```
muditchaudhary, line 305: Updating the `targetCompatibility` will be a breaking change for consumers running on an older JDK that were previously supported. It was probably bumped for supporting `Cleaner`. So, we should look at some alternative implementation there.

CedarJava/src/main/java/com/cedarpolicy/model/policy/PolicySet.java
```diff
@@ -142,6 +148,92 @@ public static PolicySet parsePolicies(String policiesString) throws InternalExce
...
+        }
+        return Optional.of(id);
+    }
+
+    private void preparseOnRustSide(String id) throws AuthException {
+        try {
+            String input = objectWriter().writeValueAsString(
+                    new PreparsePolicySetRequest(id, this));
+            String response = SharedCedarInternals.callCedarJNI("PreparsePolicySet", input);
+            JsonNode node = new com.fasterxml.jackson.databind.ObjectMapper().readTree(response);
```
muditchaudhary: Nit: Replace fully-qualified inline usage of `ObjectMapper` with import.

```
  import com.fasterxml.jackson.databind.ObjectMapper;
```

CedarJava/src/main/java/com/cedarpolicy/model/policy/PolicySet.java
```diff
@@ -142,6 +148,92 @@ public static PolicySet parsePolicies(String policiesString) throws InternalExce
...
+
+        @Override
+        public void run() {
+            SharedCedarInternals.removeCachedPolicySet(id);
+        }
+    }
+
+    @com.fasterxml.jackson.annotation.JsonInclude(com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL)
+    private static final class PreparsePolicySetRequest {
+        @com.fasterxml.jackson.annotation.JsonProperty("policySetId")
```
muditchaudhary: Nit: Replace inline with import

CedarJavaFFI/src/interface.rs
```diff
@@ -103,6 +110,46 @@ pub fn callCedarJNI(
...
+    _class: JClass<'_>,
+    j_call: JString<'_>,
+    j_input: JString<'_>,
+) -> jstring {
+    call_cedar_jni(env, j_call, j_input)
+}
+
+/// JNI entry point for NativeHelpers (shared by CachedAuthorizationEngine)
+#[jni_fn("com.cedarpolicy.NativeHelpers")]
+pub fn callCedarJNI(
```
muditchaudhary: We'll have two methods called `callCedarJNI` which are essentially the same but add methods to two different classes in Java (`BasicAuthorizationEngine` and `NativeHelpers`). `com.cedarpolicy.NativeHelpers` only exists to provide the generic JSON dispatch interface for `preParseSchema` and `preParsePolicySet`.

However, I believe it would be cleaner to just have them as direct JNI entry methods i.e., `preParsePolicySetJni` and `preParsePolicySetJni` and remove `NativeHelpers`. For example, `parseJsonSchemaJni`

CedarJavaFFI/src/interface.rs
```diff
@@ -52,6 +54,18 @@ const V0_AUTH_PARTIAL_OP: &str = "AuthorizationPartialOperation";
...
+const V0_PREPARSE_SCHEMA: &str = "PreparseSchema";
+const V0_STATEFUL_AUTH_OP: &str = "StatefulAuthorizationOperation";
+const V0_REMOVE_CACHED_POLICY_SET: &str = "RemoveCachedPolicySet";
+const V0_REMOVE_CACHED_SCHEMA: &str = "RemoveCachedSchema";
+
+thread_local! {
+    static AUTHORIZER: Authorizer = Authorizer::new();
+}
+
+static CACHED_POLICY_SETS: LazyLock<DashMap<String, PolicySet>> = LazyLock::new(DashMap::new);
```
muditchaudhary, line 64: We should have some cache eviction policy in place. Otherwise, if someone caches a lot of long-lived policy sets, we run into a risk of running out of memory. It won't be a Java OOM "graceful" error but probably a process killed error if `CACHED_POLICY_SETS` or `CACHED_SCHEMAS` grow too much. Unlikely for most use-cases but we should have it in place. Maybe give user some control over the eviction policy (e.g., TTL, maxEntries, etc.)

CedarJavaFFI/src/interface.rs
```diff
@@ -129,6 +181,216 @@ pub(crate) fn call_cedar(call: &str, input: &str) -> String {
...
+    }
+}
+
+fn json_stateful_is_authorized(input: &str) -> serde_json::Result<String> {
+    let call: StatefulAuthCall = serde_json::from_str(input)?;
+
+    // Look up cached policy set
+    let policies = CACHED_POLICY_SETS
+        .get(&call.preparsed_policy_set_id)
+        .map(|r| r.clone());
```
muditchaudhary, line 378: Maybe this could be further optimized by using `Arc<PolicySet>` instead of performing a clone as isAuthorized just takes a reference to the policy set?

@mark-creamer-amazon What do you think?

CedarJavaFFI/src/interface.rs
```diff
@@ -129,6 +181,216 @@ pub(crate) fn call_cedar(call: &str, input: &str) -> String {
...
+    // Parse principal, action, resource via FFI EntityUid type
+    let principal: cedar_policy::ffi::EntityUid = match serde_json::from_value(call.principal) {
+        Ok(p) => p,
+        Err(e) => {
+            return serde_json::to_string(&make_auth_failure(format!(
+                "failed to parse principal: {e}"
+            )));
+        }
+    };
+    let action: cedar_policy::ffi::EntityUid = match serde_json::from_value(call.action) {
```
muditchaudhary: Nit: use import for `EntityUid` instead of a fully-qualified name

CedarJavaFFI/src/interface.rs
```diff
@@ -129,6 +181,216 @@ pub(crate) fn call_cedar(call: &str, input: &str) -> String {
...
+    let resource: cedar_policy::ffi::EntityUid = match serde_json::from_value(call.resource) {
+        Ok(r) => r,
+        Err(e) => {
+            return serde_json::to_string(&make_auth_failure(format!(
+                "failed to parse resource: {e}"
+            )));
+        }
+    };
+
+    let principal = match principal.parse(Some("principal")) {
```
muditchaudhary, line 431: Wondering if we can just reuse cedar_policy::ffi::[is_authorized_json_str](https://docs.rs/cedar-policy/latest/cedar_policy/ffi/fn.is_authorized_json_str.html) or cedar_policy::ffi::[is_authorized_json_str](https://docs.rs/cedar-policy/latest/cedar_policy/ffi/fn.is_authorized_json_str.html) like we do for [non-stateful isAuthorized call](https://github.com/cedar-policy/cedar-java/blob/main/CedarJavaFFI/src/interface.rs#L116C23-L116C45) instead of doing all the request parsing by ourselves.

cedar_policy::ffi already creates a [thread local authorizer](https://github.com/cedar-policy/cedar/blame/main/cedar-policy/src/ffi/is_authorized.rs#L43). Actually, looking at the cedar_policy::ffi implementation seems like it has some version of [pre-parsed policy sets](https://docs.rs/cedar-policy/latest/cedar_policy/ffi/fn.preparse_policy_set.html) and [schemas](https://docs.rs/cedar-policy/latest/cedar_policy/ffi/fn.preparse_schema.html) methods. Maybe we can re-use it? However, from a brief look I don't see it having any eviction mechanism though.

CedarJava/build.gradle
geekphilosophy: README says that this requires JDK17 and has for 2 years. Alternative Cleaner implementation is to bring in "Janitor" from Corretto. That was actually my original implementation before seeing the README..

CedarJava/src/main/java/com/cedarpolicy/model/policy/PolicySet.java
```diff
@@ -142,6 +148,92 @@ public static PolicySet parsePolicies(String policiesString) throws InternalExce
...
+        cacheId = id;
+        SharedCedarInternals.registerCleanup(this, new PolicySetCacheCleanup(id));
+    }
+
+    /**
+     * Get the cache key for this policy set, if cached.
+     *
+     * @return The cache key if cached, or empty if not cached.
+     */
+    public Optional<String> cacheKey() {
```
muditchaudhary, line 189: The cache'd `PolicySet` can become stale if someone updates its components post `.cache()`. `PolicySet` components should have been private members and we should have used getters and setters, which would have made it easy to invalidate cache on mutation. However, making those private now would be a breaking change.

Doesn't have to be within this PR but we should add some setters for the components which invalidate cache on mutations and encourage users to use those instead. In the next major version release, we can make the members private. Probably until then, we can add documentation about this behavior.

CedarJavaFFI/src/interface.rs
geekphilosophy: Addressing.

CedarJava/src/main/java/com/cedarpolicy/model/policy/PolicySet.java
geekphilosophy: Addressing.

CedarJava/src/main/java/com/cedarpolicy/model/policy/PolicySet.java
geekphilosophy: Addressing.

CedarJava/build.gradle
geekphilosophy: Switching back to 1.8 and will fix the readme.

CedarJavaFFI/src/interface.rs
geekphilosophy: Will address.

CedarJavaFFI/src/interface.rs
geekphilosophy: Looking at adding a bounded cache sex (maxEntries). Configurable through system property, no TTL or LRU as that seems like overkill here.

CedarJavaFFI/src/interface.rs
geekphilosophy: cedar_policy::ffi does have preparse_policy_set, preparse_schema, and stateful_is_authorized with essentially the same structure. However, their cache is thread-local (RefCell<HashMap> inside thread_local!), which doesn't work for JNI: the Java thread that calls cache() is typically not the same thread that later calls isAuthorized() in a server thread pool. That's why we use a global DashMap — it's cross-thread. This was earlier feedback from a customer review.

geekphilosophy: Estimates are that a two level cache would save ~0.3% so probably not worth it..

CedarJava/src/main/java/com/cedarpolicy/model/policy/PolicySet.java
geekphilosophy: Adding documentation

CedarJava/src/main/java/com/cedarpolicy/model/policy/PolicySet.java
```diff
@@ -17,23 +17,34 @@
...
+import java.util.UUID;
 import java.util.stream.Collectors;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
 /** Policy set containing policies in the Cedar language. */
 public class PolicySet {
+    private static final String PROP_MAX_CACHED = "cedar.cache.maxPolicySets";
+    private static final int DEFAULT_MAX_CACHED = 1024;
```
muditchaudhary, line 40: Seems like we are not using `DEFAULT_MAX_CACHED` when PROP_MAX_CACHED is not set. Although, I do see 1024 being set in FFI.

CedarJava/build.gradle
geekphilosophy: Actually - previous code wasn't working with 1.8 either. Would die at runtime. After discussion with Mudit, upgraded project to 17.

CedarJava/src/jmh/java/com/cedarpolicy/AuthorizationBenchmark.java
```diff
@@ -114,28 +145,196 @@ private void setUpMediumScenario() throws Exception {
...
+    }
+
+    @Benchmark
+    public AuthorizationResponse cachedXLarge() throws AuthException {
+        return engine.isAuthorized(smallRequest, xlargePolicySetCached, smallEntities);
+    }
+
+    // --- Cross-product: policy size x entity size ---
+
+    // Small policies x {medium, large} entities
```
mark-creamer-amazon, line 251: Was wondering at first why we're leaving out the small policies x small entities case etc, then realized that yeah, they're above!

I wonder if there's an enumerating annotation we could use to have the same benchmarked method called across the parameters, similar to TestNg's `DataProvider`.

I see jmh has [`Param`](https://javadoc.io/doc/org.openjdk.jmh/jmh-core/1.27/org/openjdk/jmh/annotations/Param.html), which perhaps if we enumerated over an enum of sizes as input parameters, those could key into a map whose values are our various instance variables representing the differently sized policySet and entities.

CedarJava/src/main/java/com/cedarpolicy/BasicAuthorizationEngine.java
```diff
@@ -65,6 +68,27 @@ public BasicAuthorizationEngine() {
...
     public AuthorizationResponse isAuthorized(com.cedarpolicy.model.AuthorizationRequest q,
             PolicySet policySet, Set<Entity> entities) throws AuthException {
+        // Use the stateful (cached) path only if the policy set is cached AND
+        // any provided schema is also cached. Fall back to the uncached path
+        // otherwise to avoid silently skipping schema validation.
+        Optional<String> policySetKey = policySet.cacheKey();
+        if (policySetKey.isPresent()) {
+            boolean hasSchema = q.schema.isPresent();
+            Optional<String> schemaKey = hasSchema ? q.schema.get().cacheKey() : Optional.empty();
+            if (!hasSchema || schemaKey.isPresent()) {
```
mark-creamer-amazon, line 78: Nit: I think we could consolidate this to one if condition and predicate:

```
Optional<String> policySetKey = policySet.cacheKey();
Optional<String> schemaKey = q.schema.isPresent() ? q.schema.get().cacheKey() : Optional.empty();

if (policySetKey.isPresent() && (!q.schema.isPresent() || schemaKey.isPresent())) {
    ...
}
```

CedarJava/src/main/java/com/cedarpolicy/model/policy/PolicySet.java
```diff
@@ -142,6 +153,87 @@ public static PolicySet parsePolicies(String policiesString) throws InternalExce
...
+     *
+     * <p>For the cached path to be used during authorization, both the policy set
+     * and any associated schema must be cached. If only the policy set is cached
+     * but the schema is not, authorization will fall back to the uncached path.
+     *
+     * <p>This method is idempotent — calling it multiple times has no effect.
+     *
+     * @throws AuthException if the policies fail to parse during caching.
+     */
+    public synchronized void cache() throws AuthException {
```
muditchaudhary: We should not be throwing `AuthException` for `cache()` as `AuthException` is meant for:
> Exception thrown when errors are encountered that prevent communication with the authorization engine. 

We can introduce a new exception.

CedarJava/src/main/java/com/cedarpolicy/model/policy/PolicySet.java
```diff
@@ -142,6 +153,87 @@ public static PolicySet parsePolicies(String policiesString) throws InternalExce
...
+        if (cacheId != null) {
+            return;
```
muditchaudhary: In the current state, the user might call `cache()` on an already cached set and as they don't get any error or response about it being they might assume that the updated set is cached. They won't realize that it is stale until a stale auth request occurs or if they read the documentation carefully. We should either allow cache updates (I would prefer this as it is better CX and offers more flexibility) or throw in an exception if it is cached already (this will make the method non-idempotent).

For either of the two options, we can enforce idempotency by storing the cached input and comparing the current cache request with the already cached input. If it is the same, we don't do anything. If it is different, we update the cache or throw exception depending on the direction we go.

CedarJava/src/main/java/com/cedarpolicy/model/policy/PolicySet.java
```diff
@@ -142,6 +153,87 @@ public static PolicySet parsePolicies(String policiesString) throws InternalExce
...
+    private boolean preparseOnRustSide(String id) throws AuthException {
+        try {
+            String policiesJson = objectWriter().writeValueAsString(this);
+            preparsePolicySetJni(id, policiesJson);
+            return true;
+        } catch (JsonProcessingException e) {
+            throw new AuthException("JSON Serialization Error", e);
+        } catch (InternalException e) {
+            if (e.getMessage() != null && e.getMessage().contains("cache is full")) {
+                return false;
```
muditchaudhary: When cache is full, it looks like the subsequent cache calls will "fail" silently as upstream `cache()` method does not return any error in this case which would be confusing for the user to debug.

CedarJavaFFI/src/interface.rs
```diff
@@ -103,6 +115,159 @@ pub fn callCedarJNI(
...
+    j_id: JString<'_>,
+    j_policies_json: JString<'_>,
+) {
+    let id: String = match env.get_string(&j_id) {
+        Ok(s) => s.into(),
+        Err(_) => return,
+    };
+    let policies_json: String = match env.get_string(&j_policies_json) {
+        Ok(s) => s.into(),
+        Err(_) => return,
```
muditchaudhary: This will cause silent failures. We should expose these errors.

For example:
```
/// public string-based JSON interface to parse a schema in Cedar's cedar-readable format
#[jni_fn("com.cedarpolicy.model.schema.Schema")]
pub fn parseCedarSchemaJni<'a>(mut env: JNIEnv<'a>, _: JClass, schema_jstr: JString<'a>) -> jvalue {
    match parse_cedar_schema_internal(&mut env, schema_jstr) {
        Ok(v) => v.as_jni(),
        Err(e) => jni_failed(&mut env, e.as_ref()),
    }
}

fn parse_json_schema_internal<'a>(
    env: &mut JNIEnv<'a>,
    schema_jstr: JString<'a>,
) -> Result<JValueOwned<'a>> {
    if schema_jstr.is_null() {
        raise_npe(env)
    } else {
        let schema_jstring = env.get_string(&schema_jstr)?;
        let schema_string = String::from(schema_jstring);
        match Schema::from_json_str(&schema_string) {
            Err(e) => Err(Box::new(e)),
            Ok(_) => Ok(JValueGen::Object(env.new_string("success")?.into())),
        }
    }
}
```

CedarJavaFFI/src/interface.rs
```diff
@@ -103,6 +115,159 @@ pub fn callCedarJNI(
...
+    };
+    let policy_set_ffi: PolicySetFFI = match serde_json::from_str(&policies_json) {
+        Ok(ps) => ps,
+        Err(e) => {
+            let _ = env.throw_new(
```
muditchaudhary, line 160: I believe we can reuse `jni_failed` method here to construct the error.

CedarJavaFFI/src/interface.rs
```diff
@@ -103,6 +115,159 @@ pub fn callCedarJNI(
...
+                format!("Failed to parse policies: {msg}"),
+            );
+        }
+    }
+}
+
+/// Direct JNI entry point to remove a cached policy set by ID.
+#[jni_fn("com.cedarpolicy.model.policy.PolicySet")]
+pub fn removeCachedPolicySetJni(mut env: JNIEnv<'_>, _class: JClass<'_>, j_id: JString<'_>) {
+    if let Ok(id_str) = env.get_string(&j_id) {
```
muditchaudhary: Silent failure similar to one mentioned above when using `env.get_string(_)`

CedarJavaFFI/src/interface.rs
```diff
@@ -129,6 +295,149 @@ pub(crate) fn call_cedar(call: &str, input: &str) -> String {
...
+    };
+
+    // Parse entities
+    let entities = match CedarEntities::from_json_value(call.entities, schema.as_ref()) {
+        Ok(e) => e,
+        Err(e) => return serde_json::to_string(&make_auth_failure(e.to_string())),
+    };
+
+    // Authorize using cached data
+    let response =
```
muditchaudhary, line 471: I'm a bit skeptical of not reusing [cedar_policy::ffi::is_authorized](https://docs.rs/cedar-policy/latest/cedar_policy/ffi/fn.is_authorized.html) here. We use it through `cedar_policy::ffi::is_authorized_json_str` for our non-stateful calls. I would prefer to keep the stateful and non-stateful isAuth calls to be as same as possible when hitting `cedar_policy` to avoid potential drifts.

For example, here we set the ans to:
```
    let ans = AuthorizationAnswer::Success {
        response: response.into(),
        warnings: vec![],
    };
```
However, for non-stateful requests this is constructed like:
```
pub fn is_authorized(call: AuthorizationCall) -> AuthorizationAnswer {
    match call.parse() {
        WithWarnings {
            t: Ok((request, policies, entities)),
            warnings,
        } => AuthorizationAnswer::Success {
            response: AUTHORIZER.with(|authorizer| {
                authorizer
                    .is_authorized(&request, &policies, &entities)
                    .into()
            }),
            warnings: warnings.into_iter().map(Into::into).collect(),
        },
        WithWarnings {
            t: Err(errors),
            warnings,
        } => AuthorizationAnswer::Failure {
            errors: errors.into_iter().map(Into::into).collect(),
            warnings: warnings.into_iter().map(Into::into).collect(),
        },
    }
}
```

Although we are constructing `AuthorizationAnswer::Failure` in `make_auth_failure` to replicate this behavior but that just adds another piece of code that we'll have to keep in sync for future and introduces a potential source of inconsistency for the same API based on whether stateful or non-stateful isAuth request was made. We'll still need to use `make_auth_failure` for exposing errors for pre-parsed schema/PS failures when we send the parsed resources to `cedar_policy::ffi::is_authorized`

geekphilosophy: Agreed we should avoid drift. The immediate gap I see is that we're hardcoding warnings: vec![] in the stateful success path, dropping any parse warnings that would normally come from entity/context parsing with a schema present.

I think the fix is to propagate warnings from our context/entity parsing (matching how the upstream call.parse() collects them), while keeping make_auth_failure for the pre-lookup errors (policy set/schema not found). Does that match what you're looking for?

On the broader point of reusing cedar_policy::ffi::is_authorized directly — the blocker is that it takes an AuthorizationCall with an unparsed PolicySet (the FFI struct), so we'd have to re-serialize our cached cedar_policy::PolicySet back to the FFI representation just to have it re-parsed, which defeats the purpose of caching. If upstream ever exposes a variant that accepts a pre-resolved PolicySet + Schema we could delegate fully. Until then, is keeping the response construction in sync (especially warnings) sufficient?

muditchaudhary: > I think the fix is to propagate warnings from our context/entity parsing (matching how the upstream call.parse() collects them), while keeping make_auth_failure for the pre-lookup errors (policy set/schema not found). Does that match what you're looking for?

Yeah, we should do this and then we can iterate to reduce the surface for potential drifts

CedarJava/src/test/java/com/cedarpolicy/IntegratedCachingTest.java
```diff
@@ -0,0 +1,146 @@
...
+        assertEquals(SuccessOrFailure.Success, permitResp.type);
+        assertTrue(permitResp.success.get().isAllowed());
+
+        AuthorizationResponse denyResp = engine.isAuthorized(request(), deny, emptyEntities());
+        assertEquals(SuccessOrFailure.Success, denyResp.type);
+        assertFalse(denyResp.success.get().isAllowed());
+    }
+
+    @Test
+    public void stressCacheMany() throws AuthException, CacheException {
```
muditchaudhary, line 135: Let's also test the cache full scenario

CedarJavaFFI/src/interface.rs
mark-creamer-amazon: Yup, I think this is correct and performing the clone on the reference is cheaper.

CedarJavaFFI/src/interface.rs
muditchaudhary: Talked offline. If schema and policy sets are pre-parsed, generating the warnings here would require another parse of schema. Also, looking at stateful auth in `cedar_policy::ffi` does not populate warnings. So, there is some precedent. We can leave it as-is in this case.

muditchaudhary (approved):

mark-creamer-amazon (approved):

muditchaudhary (approved):

PR #352, opened 2026-05-11
Bugfix: EntityIdentifier equals implementation

muditchaudhary:
Fixes incorrect `equals` implementation for `EntityIdentifier` and adds tests.

### DIFF
--- a/CedarJava/src/main/java/com/cedarpolicy/value/EntityIdentifier.java
+++ b/CedarJava/src/main/java/com/cedarpolicy/value/EntityIdentifier.java
@@ -16,7 +16,6 @@
 
 package com.cedarpolicy.value;
 
-
 import com.cedarpolicy.loader.LibraryLoader;
 
 /**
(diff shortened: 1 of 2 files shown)


john-h-kastner-aws (approved):
Looks good

CedarJava/src/main/java/com/cedarpolicy/value/EntityIdentifier.java
```diff
@@ -54,8 +54,6 @@ public String toString() {
     @Override
     public boolean equals(Object o) {
         if (o == null) {
-            return true;
-        } else if (o == this) {
```
mark-creamer-amazon, line 58: Yeah I think it's fine if we skip the reference check case of `o == this`. This is captured in the below `else`.

Also a nit that here or in future releases we could switch to using `instanceof` instead of catching the `ClassCastException`, like

```
if (!(o instanceof EntityIdentifier)) return false;
```

mark-creamer-amazon (approved):

Issue #68, opened 2023-12-04
Provide a mechanism to ignore malformed policy.


yaphatak:
### Category

User level API changes

### Describe the feature you'd like to request

Provide a configuration on authorize API to Ignore malformed policy and return that malformed policyId in addition to authorization result.

### Describe alternatives you've considered

Provide an API to validate the cedar policy. This is not schema validation, but simple syntax or grammar validation.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

andrewmwells-amazon:
This seems like a fairly deep behavior change (i.e., you'd have to add a corresponding API to the Rust). Personally, I think adding an API to sanity-check the cedar policy is a better choice. Can you provide some more context on why you prefer a new configuration of `is_authorized`?

PR #367, opened 2026-08-24
Surface structured parse diagnostics from PolicySet.parsePolicies

jamesmulcahy:
## Problem

`PolicySet.parsePolicies` throws `InternalException` whose message is `format!("Internal JNI Error: {e}")` (`CedarJavaFFI/src/interface.rs`, in `jni_failed`). For a parse failure that discards everything `miette` recorded — the source span, the tokens the parser expected, the help text — and because [`ParseErrors`' `Display` prints only its first error](https://docs.rs/cedar-policy/latest/cedar_policy/struct.ParseErrors.html), every subsequent error is lost too.

A policy author writing this:

```cedar
forbid(principal, Foo::Action::"Read", resource == Foo::Table::"t");
```

sees only:

```
Internal error: Internal JNI Error: unexpected token `::`
```

No location, and no hint that the mistake is a missing `action ==`. The Rust CLI, on the same input:

```
× failed to parse policy set
╰─▶ unexpected token `::`
 ╭────
 1 │ forbid(principal, Foo::Action::"Read", resource == Foo::Table::"t");
 ·                      ─┬
 ·                       ╰── expected `!=`, `)`, `,`, `:`, `<`, `<=`, `==`, `>`, `>=`, `in`, or `is`
 ╰────
```

Measured against cedar-java 4.10.0:

```java
PolicySet.parsePolicies("forbid(principal, Foo::Action::\"Read\", resource);");
// getMessage(): Internal error: Internal JNI Error: unexpected token `::`
// getErrors():  [Internal JNI Error: unexpected token `::`]   // size 1, always
```

## Change

Everything needed already exists in the crate:

- `PolicySet::from_str` returns `ParseErrors`, which is `IntoIterator<Item = ParseError>`
- each `ParseError` implements `miette::Diagnostic`
- `cedar_policy::ffi::DetailedError` already has `impl<E: miette::Diagnostic + ?Sized> From<&E>` — the impl the validation path already uses

So `parsePoliciesJni` downcasts `ParseErrors` and throws a new `PolicyParseException extends InternalException` carrying `List<DetailedError>`, the same representation `AuthorizationEngine.validate` already returns:

```java
PolicyParseException e = ...;
e.getDetailedErrors();
// [message=unexpected token `::`,
//  sourceLocations=[SourceLabel{label="expected `!=`, `)`, `,`, `:`, `<`, `<=`, `==`,
//                   `>`, `>=`, `in`, or `is`", start=21, end=23}]]
```

Existing `catch (InternalException e)` blocks are unaffected. If building the richer exception fails for any reason, the generic path is used, so a parse error can never turn into a different kind of failure.

## Two message changes, both up for discussion

|  | before | after |
|---|---|---|
| `getMessage()` | `Internal error: Internal JNI Error: unexpected token ...` | `Internal error: unexpected token ...` |
| `getErrors()` | `[Internal JNI Error: unexpected token ...]` | `[unexpected token ...]` |

1. The `"Internal JNI Error: "` prefix is dropped — it describes the binding rather than the policy, and reading "Internal error" for an ordinary typo suggests a fault in the library rather than something the caller can fix.
2. `getErrors()` carries one entry per parse error rather than a single entry for the whole document, which is what its plural contract always implied.

Both are independent of the diagnostics and easy to drop if you'd rather keep the strings frozen — happy to revise.

## Breaking?

Not to any type or signature: `PolicyParseException` is a subclass and no existing method changes shape. Callers string-matching on the message text of a parse failure would see the two differences above.

## Testing

- 6 new tests in `PolicyParseDiagnosticsTests` — span covers the offending token, all errors reported not just the first, help text survives where Cedar supplies it, still catchable as `InternalException`, message/`getErrors()` strings pinned, valid policy sets unaffected.
- Full `CedarJava` suite: **68,723 tests, 0 failures**.
- `cargo test` in `CedarJavaFFI`: **69 passed**.

Local FFI builds used `cargo build --release --features partial-eval` for the host target rather than the `cargo zigbuild` cross-compile path, since `zig` was not available in my environment; CI exercises the normal path.

## Scope

Deliberately limited to `parsePolicies`. `Policy.parseStaticPolicy`, `Policy.parsePolicyTemplate`, `Schema.parse` and `PolicyFormatter` lose detail the same way and would be natural follow-ups — kept out here to keep the change focused, per CONTRIBUTING.

Related: #68 asks for the *id* of a malformed policy, which is adjacent but stops short of diagnostics.

### DIFF
--- a/CedarJava/src/test/java/com/cedarpolicy/PolicyParseDiagnosticsTests.java
+++ b/CedarJava/src/test/java/com/cedarpolicy/PolicyParseDiagnosticsTests.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright Cedar Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.cedarpolicy;
+
+import com.cedarpolicy.model.DetailedError;
+import com.cedarpolicy.model.exception.InternalException;
+import com.cedarpolicy.model.exception.PolicyParseException;
+import com.cedarpolicy.model.policy.PolicySet;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/** Parse failures carry Cedar's structured diagnostics, not just a flattened string. */
+public class PolicyParseDiagnosticsTests {
+
+    @Test
+    public void parseFailureCarriesSourceSpanAndExpectedTokens() {
+        // An entity literal in the action slot: the scope needs `action == ...`.
+        String src = "forbid(principal, Foo::Action::\"Read\", resource);";
+        PolicyParseException e =
+                assertThrows(PolicyParseException.class, () -> PolicySet.parsePolicies(src));
+
+        List<DetailedError> details = e.getDetailedErrors();
+        assertEquals(1, details.size());
+        DetailedError error = details.get(0);
+        assertTrue(error.message.contains("unexpected token `::`"), error.message);
+
+        assertEquals(1, error.sourceLocations.size());
+        DetailedError.SourceLabel span = error.sourceLocations.get(0);
+        // The span must cover the offending `::`, so callers can underline it.
+        assertEquals(src.indexOf("::"), span.start);
+        assertEquals(src.indexOf("::") + 2, span.end);
+        assertTrue(span.label.orElse("").contains("expected"), span.label.toString());
+    }
+
+    @Test
+    public void everyParseErrorIsReportedNotJustTheFirst() {
+        // ParseErrors' Display prints only the first error, so the flattened path reported
+        // one string for the whole document. Both accessors now carry all of them.
+        String src = "forbid(principal, Foo::Action::\"A\", resource);\n"
+                + "permit(principal, action, resource) when { 1 + };";
+        PolicyParseException e =
+                assertThrows(PolicyParseException.class, () -> PolicySet.parsePolicies(src));
+
+        assertEquals(2, e.getDetailedErrors().size());
+        assertEquals(2, e.getErrors().size());
+    }
+
+    @Test
+    public void messagesDropTheInternalJniErrorPrefix() {
+        // That prefix describes the binding, not the policy: reading "Internal JNI Error"
+        // for an ordinary typo suggests a library fault rather than a fixable mistake.
+        PolicyParseException e = assertThrows(PolicyParseException.class,
+                () -> PolicySet.parsePolicies("forbid(principal, Foo::Action::\"Read\", resource);"));
+
+        assertEquals("Internal error: unexpected token `::`", e.getMessage());
+        assertEquals(List.of("unexpected token `::`"), e.getErrors());
+    }
+
+    @Test
+    public void helpTextSurvivesWhenCedarSuppliesIt() {
+        PolicyParseException e = assertThrows(PolicyParseException.class,
+                () -> PolicySet.parsePolicies("permit(principle, action, resource);"));
+
+        DetailedError error = e.getDetailedErrors().get(0);
+        assertTrue(error.help.isPresent(), "expected help text for an invalid scope variable");
+        assertTrue(error.help.get().contains("principal"), error.help.get());
+    }
+
+    @Test
+    public void remainsCatchableAsInternalException() {
+        // PolicyParseException extends InternalException so existing callers keep working.
+        InternalException e = assertThrows(InternalException.class,
+                () -> PolicySet.parsePolicies("permit(principal, action, resource)"));
+        assertFalse(e.getErrors().isEmpty());
+    }
+
+    @Test
+    public void validPolicySetStillParses() {
+        org.junit.jupiter.api.Assertions.assertDoesNotThrow(
+                () -> PolicySet.parsePolicies("permit(principal, action, resource);"));
+    }
+}
--- a/CedarJavaFFI/src/interface.rs
+++ b/CedarJavaFFI/src/interface.rs
@@ -22,13 +22,13 @@ use cedar_policy::ffi::{
 };
 use cedar_policy::{
     ffi::{is_authorized_json_str, validate_json_str},
-    Authorizer, Entities as CedarEntities, EntityUid, Policy, PolicySet, Request, Schema, SlotId,
-    Template,
+    Authorizer, Entities as CedarEntities, EntityUid, ParseErrors, Policy, PolicySet, Request,
+    Schema, SlotId, Template,
 };
 use cedar_policy_formatter::{policies_str_to_pretty, Config};
 use dashmap::DashMap;
 use jni::{
-    objects::{JClass, JObject, JString, JValueGen, JValueOwned},
+    objects::{JClass, JObject, JString, JThrowable, JValueGen, JValueOwned},
     sys::{jstring, jvalue},
     JNIEnv,
 };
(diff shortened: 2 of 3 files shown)


jamesmulcahy:
Full disclosure -- I've not written any rust before myself, and Claude helped with this change.  Background/motivation is well summarized by Claude above.  The TL;DR is that the `cedar` CLI gives *much* better error output than the Java API -- and I'm trying to improve the experience through Java so our users can be more meaningful & actionable feedback when they provide an invalid policy.

jamesmulcahy:
Hi @lianah @mark-creamer-amazon @muditchaudhary -- James from Netflix here, we met a few weeks ago!

Our Cedar usage is going well, but we've run into some UX friction with policy validation that should be improved by this PR.  Would appreciate your time in reviewing!  Thank you!

mark-creamer-amazon:
Hey James, I'll take a look today. Thanks for the PR!

CedarJava/src/test/java/com/cedarpolicy/PolicyParseDiagnosticsTests.java
```diff
@@ -0,0 +1,102 @@
...
+    public void remainsCatchableAsInternalException() {
+        // PolicyParseException extends InternalException so existing callers keep working.
+        InternalException e = assertThrows(InternalException.class,
+                () -> PolicySet.parsePolicies("permit(principal, action, resource)"));
+        assertFalse(e.getErrors().isEmpty());
+    }
+
+    @Test
+    public void validPolicySetStillParses() {
+        org.junit.jupiter.api.Assertions.assertDoesNotThrow(
```
mark-creamer-amazon, line 99: Nit: we should import `assertDoesNotThrow` above

mark-creamer-amazon:
I agree with the direction of this PR, but I do worry about breaking existing consumers here.

1. Removing the `Internal JNI Error:` prefix could of course technically break a consumer inspecting and branching on this prefix's presence (unless they strip it).
   * I think this is the smaller risk I'd personally be willing to move forward regardless
2. Similarly I've seen at least one downstream consumer that is doing a strict regex match on `getMessage()`, which would break with this change, e.g. `java.util.regex.Pattern.compile("^Internal error: Internal JNI Error: unexpected token `([^`]*)`$");`. More commonly there's likely other consumers that branch based on what's expected to be a single message, e.g.
    ```
    if (e.getMessage().equals("Internal error: Internal JNI Error: failure a")) {
        ...
    } else if (e.getMessage().equals("Internal error: Internal JNI Error: failure b")) {
        ...
    }
    ```
    I think this class of issue class is enough where 
3. While `getErrors` does return a `List` and implies plurality, all `InternalException`'s from the FFI thus far have been using the [basic constructor](https://github.com/cedar-policy/cedar-java/blob/main/CedarJava/src/main/java/com/cedarpolicy/model/exception/InternalException.java#L28), which only populates one element. But as `getErrors()` returns a `List`, I'll concede that perhaps it was never safe for a consumer to expect `.size() == 1` or refer solely to `errs.get(0)` as the only element. I personally think that us fully populating the `getErrors()` result list is the right direction.

I think I'm hesitant towards the `getMessage()` behavior change, but I'm fine with `getErrors()` changing to return the other errors.

jamesmulcahy:
> I agree with the direction of this PR, but I do worry about breaking existing consumers here.

[...]

> I think I'm hesitant towards the `getMessage()` behavior change, but I'm fine with `getErrors()` changing to return the other errors.

Understood, I think that's a reasonable take.  I'll update the PR to conform with your guidance.  Thanks for the review!

CedarJava/src/test/java/com/cedarpolicy/PolicyParseDiagnosticsTests.java
jamesmulcahy: Ack; fixed this in my local branch

jamesmulcahy:
@mark-creamer-amazon — Claude Code here, posting as James's agent; he's asked me to follow up on your review and has reviewed this reply before it went out.

Two commits pushed, branch now at `ffdf01b`.

### `getMessage()` is unchanged

Point 2 taken in full — the message is now byte-for-byte what the generic error path produced, prefix and all. Your example pattern

```java
Pattern.compile("^Internal error: Internal JNI Error: unexpected token `([^`]*)`$");
```

still matches, and it can't be widened by a document with several errors: `ParseErrors`' `Display` prints only the first error, and the message is still exactly that, so populating the list doesn't leak into the string. There's a test pinning the literal value now (`messageIsUnchangedForBackCompat`) — it asserts the inverse of what the earlier revision asserted, so the regression you flagged is now guarded rather than introduced.

Mechanically: `PolicyParseException` takes the message as a constructor argument, and the Rust side has a single `internal_error_message()` helper used by both the enriched path and the plain `InternalException` fallback, so the two can't drift apart. `InternalException` gained a `protected` constructor that sets message and error list independently — the existing constructors derive the message by joining the list, which is what made a per-error list necessarily widen the message. Existing constructors are untouched.

### The one point I'd like your explicit call on: the prefix on `getErrors()` entries

`getErrors()` now returns one entry per parse error, per your point 3. What's still a judgement call is what each entry contains. As pushed:

```
getMessage() -> "Internal error: Internal JNI Error: unexpected token `::`"   (unchanged, prefix retained)
getErrors()  -> ["unexpected token `::`", "unexpected token `}`"]             (bare, no prefix)
```

To be explicit: the original error message keeps the prefix exactly where it has always been. The prefix is absent only from the individual list elements. The reasoning is that `"Internal JNI Error: "` describes the binding rather than any one parse error, so once the list is per-error, repeating it on every element reads as though each error were its own separate JNI failure.

Your point 1 said you'd accept removing the prefix as the smaller risk, but that was said about `getMessage()`, and I don't want to assume it carries over to the list. The conservative alternative is to prefix element 0 only — preserving the exact former value of `getErrors().get(0)` for anyone reading just that element, leaving 1..n bare. Strictly non-breaking, though incoherent as a contract.

**Does the bare-element version match your expectations, or would you prefer element 0 keep the prefix?** Either is fine by us; it's a small change and we'd rather have your explicit sign-off than guess.

### Also in this push

* `getDetailedErrors()` returns Cedar's `miette` diagnostics per error — message, source span, expected tokens, help text. This is where the new information lives now that the message is fixed.
* Javadoc on `DetailedError.SourceLabel` documenting that the spans are **UTF-8 byte offsets**, not `String` indices. Worth calling out because the obvious `source.substring(start, end)` throws `StringIndexOutOfBoundsException` on any policy text containing a non-ASCII character, which we hit while testing. It also records that offsets are absolute within the whole parsed text rather than per-policy, that a span can be empty (an unterminated string literal produces one), and that the span covers the unexpected token — which for a missing operand is the token that followed it, sometimes a line later.
* `assertDoesNotThrow` import nit from your inline comment.
* A `checkstyleMain` whitespace violation this PR had introduced in `PolicyParseException` — that was failing the build and I should have caught it earlier.

Verified locally: the 6 `PolicyParseDiagnosticsTests` pass, `javadoc` and `checkstyleMain` are clean, and the full suite's failure count is identical to the pre-change baseline on this tree (58, all in `SharedIntegrationTests` and unrelated to parsing — happy to share the before/after if useful).

Thanks again for the careful review — the `getMessage()` objection was the right call and the change is better for it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jamesmulcahy:
@mark-creamer-amazon I pushed a fix for the CI failure; if you can approve again I'm optimistic we'll get a clean run, but I'll keep an eye on it!

CedarJava/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Added Offset function support [#331](https://github.com/cedar-policy/cedar-java/pull/331)
 * Added PolicySet to JSON conversion API [#329](https://github.com/cedar-policy/cedar-java/pull/329)
 * Added Cedar Schema support for Entity Validation [#332](https://github.com/cedar-policy/cedar-java/pull/332)
+* Added `PolicyParseException`, thrown by `PolicySet.parsePolicies` when policy text fails to parse. It is a subclass of `InternalException`, so existing `catch` blocks and `getMessage()` are unaffected, and adds `getDetailedErrors()` returning Cedar's structured diagnostics - source span, expected tokens, and help text - for each error. `getErrors()` now carries one entry per parse error rather than a single entry for the whole document [#367](https://github.com/cedar-policy/cedar-java/pull/367)
```
muditchaudhary, line 11: Note to myself: CHANGELOG.md in `main` is stale. The features under "Unreleased" are now released in CedarJava 4.8. I will update it and also place this change under the correct section.

CedarJavaFFI/src/interface.rs
```diff
@@ -534,6 +534,88 @@ struct JavaInterfaceCall {
...
+/// richer exception and the fallback below can never drift apart.
+fn internal_error_message(errs: &ParseErrors) -> String {
+    format!("Internal JNI Error: {errs}")
+}
+
+/// Throw a plain `InternalException`, exactly as `jni_failed` would have.
+fn throw_internal(env: &mut JNIEnv<'_>, errs: &ParseErrors) {
+    // We have to unwrap here as we're doing exception handling
+    // If we don't have the heap space to create an exception, the only valid move is ending the process
+    env.throw_new(
```
muditchaudhary, line 625: `throw_internal` can be reached with a Java exception already pending: every `? `in `build_parse_exception` leaves one parked when it fails, although this is rare (E.g., new_object -> `NoClassDefFoundError` if the `.jar` predates the `.so`, `new_string/new_object_array` -> `OutOfMemoryError`). 

In these cases `throw_new` then returns `Err(Error::JavaException)` because the crate correctly refuses to call JNI in that state, and `.unwrap()` panics which I believe would abort the JVM.

One fix would be to reuse `jni_failed`, which already guards on `exception_check()` and formats the identical message or add `env.exception_check().unwrap_or_default()` check here.


Rest of the PR looks good to me. Thank you for working on this!

CedarJava/CHANGELOG.md
muditchaudhary: Regarding removing the prefix in `getErrors`, it seems fine to me. @mark-creamer-amazon what do you think? IIRC Cedar does not treat mutating error messages as a breaking change.

muditchaudhary (changes requested):

CedarJavaFFI/src/interface.rs
jamesmulcahy: Good catch — you're right, and thank you for tracing it so precisely. Fixed in 129ef7d.

Confirming the chain for the record: each `?` in `build_parse_exception` bottoms out in a jni-rs call whose `jni_non_null_call!`/`jni_non_void_call!` macro ends in `check_exception!`, which returns `Error::JavaException` while leaving the exception *pending* — it detects, it doesn't clear. `throw_internal` then resolves its class argument through `Desc::lookup` → `find_class`, which hits that same check and returns `Err` before `ThrowNew` is ever reached. The `.unwrap()` panics, and since `jni_fn` expands `parsePoliciesJni` to an `extern "C"` function, unwinding across that boundary aborts the process. So a recoverable `NoClassDefFoundError` from a `.jar`/`.so` version skew — the realistic trigger — took the JVM down.

I went with the `exception_check()` guard inside `throw_internal` rather than reusing `jni_failed`, because `jni_failed`'s signature isn't quite the right fit here: it takes `&dyn Error` and returns a `jvalue`, neither of which this call site wants (it's already holding a `&ParseErrors`, and it's on a path that returns `()`), so routing through it needed more adaptation at the call sites than the guard itself costs. Putting the guard in `throw_internal` rather than at the two call sites also covers the `env.throw(exception).is_err()` branch — benign today, since `throw` on an already-constructed `JThrowable` does no checked lookup, but it would be a latent instance of the same bug. Entirely happy to switch to `jni_failed` if you'd prefer the single implementation; it's a small change either way.

One note on the resulting behaviour, which I think is a slight improvement beyond the crash fix: returning early means the pending `OutOfMemoryError`/`NoClassDefFoundError` propagates to the caller when the native method returns, rather than being replaced by an `InternalException` whose message names a parse error. The original error is much more diagnostic for that class of failure. I've written that reasoning into the comment on the guard, along with why the `unwrap` below it is sound once the guard is in place — the existing "only valid move is ending the process" comment was inherited from `jni_failed`, where the guard is precisely what justifies it.

### Testing

- `cargo test --features partial-eval`: 71 passed, 0 failed.
- `cargo fmt --check`: clean. The two build warnings (`unused variable: warnings`, `method get is never used`) are pre-existing on `main`.
- `./gradlew check`: clean, SpotBugs and Checkstyle included.
- Full `CedarJava` suite: 68,030 tests, 58 failures — all in `SharedIntegrationTests`, all the same `unable to find an applicable action given the policy scope constraints` assertion. I re-ran with this commit reverted and got an identical 58, and the branch touches no corpus resources or validation code, so they're pre-existing in my environment rather than anything from this PR. (My earlier "68,723 tests, 0 failures" was from an environment where that corpus resolved differently; CI is the authority here.)

Same local-build caveat as before: host `cargo build --release --features partial-eval` with the `.so` staged manually, as `zig` isn't available to me for the `zigbuild` cross-compile path.

I'll leave the CHANGELOG placement to you as you noted, and the `getErrors()` prefix question with you and @mark-creamer-amazon — happy to restore the prefix if you'd rather keep those strings frozen.

jamesmulcahy: Yikes, Claude's just *way* too verbose here -- my apologies. I'll vet it more closely next time.  I asked it to mention two things, and it's written a thesis 🤦

muditchaudhary (approved):

CedarJava/CHANGELOG.md
mark-creamer-amazon: Yeah I think removing the prefix in each of `getErrors` should be fine.

mark-creamer-amazon (approved):

jamesmulcahy:
Thank you both! 🙇

muditchaudhary:
Thanks @jamesmulcahy for your contribution. I will include this change in the upcoming CedarJava 4.11 release in the coming week or so.

Issue #364, opened 2026-07-13
Type-aware Partial Evaluation (TPE)


muditchaudhary:
### Category

User level API changes

### Describe the feature you'd like to request

Provide [TPE](https://docs.rs/cedar-policy/latest/cedar_policy/struct.PolicySet.html#method.tpe) support in CedarJava.

### Describe alternatives you've considered

N/a

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #370, opened 2026-09-16
feat(tpe): add type-aware partial evaluation entity model and validation

muditchaudhary:
## Summary

Adds the entity model for Cedar's type-aware partial evaluation (TPE) to CedarJava: entities whose attributes, parents, and tags may each be independently unknown. Everything is `@Experimental` and purely additive — no existing API changes.

This is the entity half of the foundation. The TPE request/response types and the authorization call itself follow in a separate PR.

## What

**CedarJava**
- `PartialEntity` — EUID always known; attrs, parents, and tags each either absent (unknown) or present and complete.
- `PartialEntities` — the collection, carrying the two checks that only make sense over the whole set: no duplicate UIDs, and no in-collection parent whose own parents are unknown.
- `PartialEntityUID` — partial counterpart to `EntityUID`, reusing `EntityTypeName` and `EntityIdentifier`. The type is always known, since TPE needs it to type check a request.
- Constructors take a `Schema` and validate eagerly, so a value in hand has already been checked.

**CedarJavaFFI**
- New `tpe` Cargo feature and `src/tpe.rs` with two validators, accepting a schema in either Cedar or JSON format.

## Why

TPE evaluates policies against incomplete data, which needs a way to say "this entity exists but I don't know its attributes" — distinct from "I don't know whether this entity exists at all," which is expressed by leaving it out of the collection.

Validation happens at construction rather than at authorize time so that a bad entity fails where the mistake was made, with a message naming the offending attribute, instead of surfacing later as an opaque error from a call that looks unrelated.

Issue: #364


CedarJava/src/main/java/com/cedarpolicy/model/entity/PartialEntity.java
```diff
@@ -0,0 +1,262 @@
...
+    /**
+     * A debug rendering, laid out like {@link Entity#toString()}. Nothing in the translation layer reads it: the wire
+     * form is produced by {@link com.cedarpolicy.serializer.PartialEntitySerializer} from the getters. A field that is
+     * unknown is labelled as such, while a field that is known to be empty is omitted, the same as on a concrete
+     * entity.
+     *
+     * @return a debug rendering of this partial entity
+     */
+    @Override
+    public String toString() {
```
victornicolet, line 241: Should we have some tests for this?

CedarJava/src/test/java/com/cedarpolicy/PartialEntitiesTests.java
```diff
@@ -0,0 +1,236 @@
...
+        nullFieldsJson.putNull("parents");
+        nullFieldsJson.putNull("tags");
+        var nullFields = PartialEntities
+                .fromJson(JsonNodeFactory.instance.arrayNode().add(nullFieldsJson), TPE_SCHEMA)
+                .getEntities()
+                .iterator()
+                .next();
+        assertTrue(nullFields.getAttrs().isEmpty());
+        assertTrue(nullFields.getParents().isEmpty());
+        assertTrue(nullFields.getTags().isEmpty());
```
victornicolet, line 205: What about a test checking that a present-but-empty attrs/parents/tags in the JSON comes back as present-and-empty rather than unknown? The blocks above only cover the omitted and null cases, which map to unknown.

CedarJava/src/main/java/com/cedarpolicy/model/entity/PartialEntities.java
```diff
@@ -0,0 +1,178 @@
...
+        } catch (JsonProcessingException e) {
+            throw new InternalException("Failed to serialize the partial entities: " + e.getMessage());
+        }
+        validate(entitiesJson, schema);
+        this.entities = Set.copyOf(entities);
+    }
+
+    /**
+     * Constructs a collection from concrete entities, each of which is fully known. Mirrors
+     * {@code PartialEntities::from_concrete}.
```
victornicolet: I am not sure this is true, it calls lift, which calls from_json_value?

CedarJava/src/test/java/com/cedarpolicy/PartialEntitiesTests.java
```diff
@@ -0,0 +1,236 @@
...
+
+        // Stating that `admins` has no parents is also fine: the hierarchy can be closed.
+        var noParents =
+                new PartialEntity(admins, Optional.empty(), Optional.of(Set.of()), Optional.empty(), TPE_SCHEMA);
+        var withParent = new PartialEntities(Set.of(child, noParents), TPE_SCHEMA);
+        assertEquals(2, withParent.getEntities().size());
+    }
+
+    @Test
+    public void testRejectsDuplicateUid() throws InternalException {
```
victornicolet, line 101: This covers the set entities constructed with `Set` but maybe we should also test `from_json` (i.e. build the json array of entities).

CedarJava/src/main/java/com/cedarpolicy/model/entity/PartialEntities.java
```diff
@@ -0,0 +1,178 @@
...
+    /**
+     * Constructs a collection from a given Set of PartialEntity objects, checking the collection as a whole against the
+     * schema.
+     *
+     * @param entities The partially known entities.
+     * @param schema   The schema to check the entities against.
+     * @throws InternalException                   If the entities do not check out against the schema, or if they cannot
+     *                                             be serialized.
+     * @throws MissingExperimentalFeatureException If the native library was built without the {@code tpe} feature.
+     * @throws NullPointerException                If the schema is null.
```
mark-creamer-amazon, line 77: Are these applicable any more?

Same with below

muditchaudhary: Still applicable. `validate(...)` uses `Objects.requireNonNull(schema, ...)` which can throw `NullPointerException` and ends by routing any native error to `MissingExperimentalFeatureException`. `MissingExperimentalFeatureException` is a subclass of `InternalException`.

CedarJava/src/test/java/com/cedarpolicy/PartialEntitiesTests.java
victornicolet: Do you plan to add those tests? Or maybe I missed them? I only see the doc comment about 'from_concrete' has been updated.

muditchaudhary: I added `testFromJsonRejectsDuplicateUid` and `testFromJsonRejectsEntityThatDoesNotConformToSchema` which should cover it.

CedarJava/src/test/java/com/cedarpolicy/PartialEntitiesTests.java
```diff
@@ -0,0 +1,269 @@
...
+import static com.cedarpolicy.TestUtil.buildEuidObject;
+import static com.cedarpolicy.TestUtil.buildUidObject;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/** Tests for {@link PartialEntities}, a collection of partially known entities. */
+public class PartialEntitiesTests {
+    /** See {@link PartialEntityTests} for what this schema is shaped to catch. */
+    private static final Schema TPE_SCHEMA = TestUtil.loadSchemaResource("/tpe_schema.json");
```
mark-creamer-amazon, line 49: Maybe this matches the existing pattern, but should this be `TPE_SCHEMA_JSON`?

victornicolet (approved):

CedarJava/src/test/java/com/cedarpolicy/PartialEntitiesTests.java
muditchaudhary: Yeah, it matches the existing pattern in other test suites where we use `*_SCHEMA` for JSON schemas as `*_SCHEMA_CEDAR` for Cedar schemas

CedarJavaFFI/src/tests.rs
```diff
@@ -1035,4 +1035,143 @@ mod partial_authorization_tests {
...
+        entity Photo;
+        action view appliesTo {
+            principal: [User],
+            resource: [Photo],
+            context: { "authenticated": Bool }
+        };
+    "#;
+
+    fn schema_json() -> String {
+        json!(SCHEMA).to_string()
```
mark-creamer-amazon, line 1056: Nit: Could be worth being consistent with `interface.rs` above which uses `serde_json::json!(SCHEMA).to_string()`. Or perhaps the `SCHEMA` could be shared in both

CedarJava/src/main/java/com/cedarpolicy/model/entity/PartialEntity.java
```diff
@@ -0,0 +1,262 @@
...
+                schema);
+    }
+
+    /**
+     * Construct a partial entity without checking it against a schema. Only for callers that go on to have Cedar check
+     * the whole collection in a single native call, which subsumes the per-entity checks and additionally covers the
+     * rules that span the collection. In practice that means {@link PartialEntities}, on both the JSON and the concrete
+     * path.
+     */
+    PartialEntity(EntityUID euid, Optional<Map<String, Value>> attrs, Optional<Set<EntityUID>> parents,
```
mark-creamer-amazon, line 114: Nit: should we add the explicit `public` here?

muditchaudhary: I deliberately did not make it `public` here as it is used to create non-Schema validated `PartialEntity` which should not happen (for consumers). This method exists merely to be used by `PartialEntities` constructors to avoid duplicate validations. So, keeping it package-private

mark-creamer-amazon (approved):

